PR #7, opened 2022-06-21
Dockerize the Application

### LINKED ISSUE #4: Local environment deployment configuration
In order to test the application locally with ease, it would be great to have a Docker Compose description which would:
 - start a PostgreSQL database
 - start the application when the database is ready

The application should be started from a container. The DB port should be accessible to be able to connect a SQL client to browse / query it.

dheerajodha:
Closes #4

This PR mainly includes (inside the location `/src/main/docker/`):
* `Dockerfile`: to create an image of our Spring Boot Application.
* `create_tables.sql`: to create both the required database and table.
* `docker-compose.yml`: main file where we define definition to containerize both the DB and Spring Boot Application.
* `application.properties`: where we replaced the previously used encryption mechanism using `jasypt` library with the environment variables.

Steps to locally setup the application:
1. `$ mvn clean package -DskipTests`
2. `$ cp target/*.jar src/main/docker/`
3. `$ docker-compose up`

Steps to test if the application works properly:
1. `$ docker exec -it <container name of the DB> psql <database name> -U <username> -c 'select * from plugins;'`
2. You should see a table with column names 'id', 'name', 'release_timestamp', and 'scm', whose rows are filled with plugin details.


.gitignore
```diff
@@ -3,3 +3,4 @@ HELP.md
 target
 !.mvn/wrapper/maven-wrapper.jar
 !.mvn/wrapper/maven-wrapper.properties
+.env
```
ADI10HERO, line 6:
```suggestion
.env

```

ADI10HERO:
Hey @dheerajodha 
Apart from the "adding a new line at the end of file" changes I had a question...
Why are we building the project outside docker and then copying the docker file inside the container? Wouldnt it be better to build the project inside the container itself?

Also, "ideally" we shouldn't be running any other steps and just `docker compose up` should do the trick.. So what I am trying to say here is, 
```sh
$ mvn clean package -DskipTests
$ cp target/*.jar src/main/docker/
``` 
can be done inside the Dockerfile.

Let me know if I am missing something or if the current approach is how it was decided in meetings. :) 


PS: I think you can modify your IDE settings to automatically include that new line at the end of every file so that GitHub doesn't give that annoying missing symbol.

src/main/docker/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+    image: 'docker-spring-boot-postgres:latest'
+    build:
+      context: .
+    container_name: app
+    env_file:
+      - .env
+    environment:
+      - SPRING_DATASOURCE_URL=jdbc:postgresql://db:${POSTGRES_PORT}/${POSTGRES_DB}
+      - SPRING_DATASOURCE_USERNAME=${POSTGRES_USERNAME}
+      - SPRING_DATASOURCE_PASSWORD=${POSTGRES_PASSWORD}
```
alecharp, line 13: You don't need those. `POSTGRES_USERNAME`, `POSTGRES_PASSWORD` and others should be in the `.env` file already.

```suggestion
```

src/main/docker/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+  db:
+    image: postgres:14.1-alpine
+    restart: always
+    env_file:
+      - .env
+    environment:
+      - POSTGRES_USER=${POSTGRES_USERNAME}
+      - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
+    expose:
+      - 5432
```
alecharp, line 26: Because the `Dockerfile` of the image is already declaring this (https://github.com/docker-library/postgres/blob/2f6878ca854713264ebb27c1ba8530c884bcbca5/14/bullseye/Dockerfile#L221), we don't need this here. 

However, we need to say on which port the database can be reach locally to be able to connect a SQL client and query it. This would ease the development and as the Compose will not be used for production deployment, this is not a concerns for the future.

```suggestion
```

.gitignore
alecharp, line 6: You can ignore this.

```suggestion

### Application
.env
```

src/main/docker/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: '3.3'
+services:
+  app:
+    image: 'docker-spring-boot-postgres:latest'
+    build:
+      context: .
```
alecharp, line 6: Once you moved the `docker-compose.yml` from the current location.

```suggestion
      context: .
      dockerfile: src/main/docker/Dockerfile
```

You could incorporate the argument `JAR_FILE` from the current state of the `Dockerfile` (see  https://docs.docker.com/compose/compose-file/build/).

src/main/docker/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: '3.3'
+services:
+  app:
+    image: 'docker-spring-boot-postgres:latest'
```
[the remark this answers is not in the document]
dheerajodha: Oh right, we're already creating a custom image using the Dockerfileso we won't need any other ready-to-use image at all.

src/main/docker/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+    build:
+      context: .
+    container_name: app
```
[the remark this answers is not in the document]
dheerajodha: You are suggesting removing this line, and instead rename this container by replacing it's current name [on line 3](https://github.com/jenkins-infra/plugin-health-scoring/blob/1cbfcd8698289151239f52155d5ce251eb8b6d6a/src/main/docker/docker-compose.yml#L3) with `<docker_compose_context>_<service_name>` format?

If yes, then can you tell me more about "docker compose context" and "service name"?

Would it be something like `backend_app` ...?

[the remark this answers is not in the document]
alecharp: I meant that we don't need this configuration. We can keep the default naming. The `docker_compose_context` is the application name in the sense of Docker Compose: it's the name of the folder you are running Docker Compose from. The service name is the name you chose line 3 (https://github.com/jenkins-infra/plugin-health-scoring/pull/7/files#diff-c3d30d4845cc6f240782a49f91ce0c35378d163006beee6530bf3448719db99bR3)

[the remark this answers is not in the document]
dheerajodha: Understood :+1:

src/main/docker/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+    restart: always
+    env_file:
+      - .env
+    environment:
+      - POSTGRES_USER=${POSTGRES_USERNAME}
+      - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
+    expose:
+      - 5432
+    volumes: 
+      - ./create_tables.sql:/docker-entrypoint-initdb.d/create_tables.sql
```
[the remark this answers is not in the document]
dheerajodha: [Correct me if I'm wrong] I understand your point now, "volumes" includes a SQL script that creates a DB and an empty table. This script will be used every time we use `docker-compose up`, meaning, on every startup, an empty DB gets created, which is an overkill.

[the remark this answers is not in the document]
alecharp: it's overkill, yes and no. Here, we can work without this. Also, you step hibernate to create the tables required by the objects (see line 7-8 of `application.properties`).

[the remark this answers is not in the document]
dheerajodha: Yes exactly, hence there's no point in having this script at all.

docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+      db:
+        condition: service_healthy
+
+  db:
+    image: postgres:14.1-alpine
+    restart: always
+    env_file:
+      - .env
+    healthcheck:
+      test: ["CMD", "pg_isready", "-h", "${POSTGRES_HOST}", "-p", "5432", "-U", "${POSTGRES_USER}", "-d", "${POSTGRES_DB}"]
```
dheerajodha, line 18: We were seeing the error "service unhealthy", while docker-compose was setting up the database because I mistakenly wrote "-u" instead of "-U". This caused healthcheck to fail every time with an invisible "unknown flag" command I think.

src/main/docker/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM openjdk:17-oracle
+RUN groupadd -r phs && useradd -r -g phs phs
+USER phs:phs
+ARG DEPENDENCY=./target/dependency
+COPY ${DEPENDENCY}/BOOT-INF/lib /app/lib
+COPY ${DEPENDENCY}/META-INF /app/META-INF
+COPY ${DEPENDENCY}/BOOT-INF/classes /app
+ENTRYPOINT ["java","-cp","./target/dependency/BOOT-INF/lib:/app/lib/*","io.jenkins.pluginhealth.scoring.PluginHealthScoring"]
```
dheerajodha, line 8: I'm currently trying to understand what to put before the ":/app/lib/*" in the third element inside this array. I'll do some googling on this.

docker-compose.yml
alecharp: well done!

src/main/docker/Dockerfile
alecharp: Please see https://github.com/jenkins-infra/plugin-health-scoring/pull/7/commits/0486ea82807c5145f8ba09896198b560f5716c5e

You still need to `mvn package` before you can `docker image build` it's working now.

alecharp:
We still need to document how anyone can run the application locally.

dheerajodha:
@alecharp I see this error when I run "mvn spring-boot:run" 

`java.lang.RuntimeException: Driver org.postgresql.Driver claims to not accept jdbcUrl, jdbc:postgresql://${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}`

Do you see it too?

alecharp:
@dheerajodha yes, this is normal. When you run `mvn spring-boot:run`, unless you have `export`ed the environment variables used in this setup, the java application is not able to understand those parameters. 

you need to run `export POSTGRES_HOST=localhost` and all the others. We could also create a profile so it's using a different configuration when in development.

dheerajodha:
@alecharp Thanks, I exported the variables and it worked. I thought about it and it made sense because I'm directly running the spring application and application.properties files have these variables whose values are inside the .env file which is fetched when we run the docker-compose command. So, the jdbc URL cannot be constructed without explicitly exporting the values.

1.) Is, creating a profile, part of this PR?

2.) Also, when I run `docker-compose up` I get this error:
`plugin-health-scoring-db-1 ... FATAL:  password authentication failed for user "${POSTGRES_USERNAME}"`
`plugin-health-scoring-db-1  ... DETAIL:  Role "${POSTGRES_USERNAME}" does not exist.`

`plugin-health-scoring-app-1  | org.postgresql.util.PSQLException: FATAL: password authentication failed for user "${POSTGRES_USERNAME}"`
I searched for the variable "POSTGRES_USERNAME" to see where it is picking up this variable but couldn't find any occurrence. I tried to generate a new jar file using `mvn package -DskipTests` but still got this error. Any ideas/hints?

alecharp:
> 1.) Is, creating a profile, part of this PR?

No.

> 2.) Also, when I run docker-compose up I get this error:
plugin-health-scoring-db-1 ... FATAL: password authentication failed for user "${POSTGRES_USERNAME}"
plugin-health-scoring-db-1 ... DETAIL: Role "${POSTGRES_USERNAME}" does not exist.

Normally, it's `POSTGRES_USER` not `POSTGRES_USERNAME`. It means that you are still referencing `*_USERNAME`  somewhere.

Look into your `.env` file.

README.md
```diff
@@ -7,3 +7,21 @@ This project aims to introduce a metric system to calculate the health score of
...
+
+**Steps**
+- Clone this repository using:
+  - *git clone https://github.com/jenkins-infra/plugin-health-scoring.git*
+- Enter inside the newly cloned repo, using:
+  - *cd plugin-health-scoring*
+- Create a JAR file of this project inside the directory *./target/*, using:
+  - *mvn package -DskipTests*
+- Run the docker-compose.yml file to start both the services, using:
+  - *docker-compose up*
```
alecharp, line 27:
```suggestion
  - *docker compose up*
```

README.md
```diff
@@ -7,3 +7,21 @@ This project aims to introduce a metric system to calculate the health score of
...
+
+And in order to start up these services and keep them up and running, we'll use a containerized PostgrSQL DB and build an image for the Spring Boot App using a custom Dockerfile. All of this will be done via docker-compose.yml file.
+
+**Steps**
+- Clone this repository using:
+  - *git clone https://github.com/jenkins-infra/plugin-health-scoring.git*
+- Enter inside the newly cloned repo, using:
+  - *cd plugin-health-scoring*
+- Create a JAR file of this project inside the directory *./target/*, using:
+  - *mvn package -DskipTests*
```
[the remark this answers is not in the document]
alecharp: For a quick build, this is fine.

README.md
alecharp: Please, don't forget to mention the `.env` files and include an example.

dheerajodha: Yes sure, just pushed. Please check, thanks!

alecharp:
Sorry for the forced push. We should never do that. I just forgot to rewrite the history locally before pushing the code.

alecharp (approved):

PR #17, opened 2022-07-04
Prevent duplication when storing plugins

### LINKED ISSUE #8: Plugins are duplicated in the database
If you start the application multiple time, you can see that the plugins are saved as many times in the database. This is not correct as all plugins are uniquely identified with their name.

- [ ] Restarting the application is not duplicating the entries in the database
- [ ] When the `SCM` or the `releaseTimestamp` are updated in the update-center, those values are updated in the database

dheerajodha:
Closes #8 

* Now when I start the app, it doesn't populate the DB with duplicate plugins.
* Now there's a check added before we save an object into the DB. The flow now first checks if for a given plugin received from the DB has the same name as any other plugin stored already inside the DB,
  * If no, then save the object in DB.
  * If yes, check if there is any difference in `releaseTimestamp` and `scm` field values, if yes (for both or any one field) then update the object present inside the DB.

### DIFF
--- a/src/main/java/io/jenkins/pluginhealth/scoring/config/DatabaseConfiguration.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/config/DatabaseConfiguration.java
@@ -22,14 +22,16 @@
  * SOFTWARE.
  */
 
-package io.jenkins.pluginhealth.scoring;
+package io.jenkins.pluginhealth.scoring.config;
 
-import org.springframework.boot.SpringApplication;
-import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.domain.EntityScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
 
-@SpringBootApplication(scanBasePackages = "io.jenkins.pluginhealth.scoring")
-public class PluginHealthScoring {
-    public static void main(String[] args) {
-        SpringApplication.run(PluginHealthScoring.class, args);
-    }
+@Configuration
+@EnableTransactionManagement
+@EnableJpaRepositories(basePackages = "io.jenkins.pluginhealth.scoring.repository")
+@EntityScan(basePackages = "io.jenkins.pluginhealth.scoring.model")
+public class DatabaseConfiguration {
 }
--- a/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
@@ -39,7 +39,7 @@ public class PluginService {
         this.pluginRepository = pluginRepository;
     }
 
-    public void updateDatabase(Plugin plugin) {
+    public void saveOrUpdate(Plugin plugin) {
         Optional<Plugin> pluginFromDatabase = pluginRepository.findByName(plugin.getName());
 
         if (pluginFromDatabase.isPresent()) {
@@ -55,10 +55,6 @@ public class PluginService {
         else {
             saveOrUpdate(plugin);
         }
-    }
-
-    public void saveOrUpdate(Plugin plugin) {
-        pluginRepository.save(plugin);
     }
 
 }
--- a/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
@@ -40,19 +40,9 @@ public class PluginService {
     }
 
     public Plugin saveOrUpdate(Plugin plugin) {
-//        return pluginRepository.findByName(plugin.getName())
-//            .map(pluginFromDatabase -> pluginFromDatabase.setScm(plugin.getScm()).setReleaseTimestamp(plugin.getReleaseTimestamp()))
-//            .map(pluginRepository::save)
-//            .orElse(pluginRepository.save(plugin));
-        if (pluginRepository.findByName(plugin.getName()).isPresent()) {
-            Optional<Plugin> pluginFromDB = pluginRepository.findByName(plugin.getName());
-            pluginFromDB.get().setScm(plugin.getScm()).setReleaseTimestamp(plugin.getReleaseTimestamp());
-            pluginRepository.save(pluginFromDB.get());
-            return pluginFromDB.get();
-        }
-        else {
-            pluginRepository.save(plugin);
-            return plugin;
-        }
+        return pluginRepository.findByName(plugin.getName())
+            .map(pluginFromDatabase -> pluginFromDatabase.setScm(plugin.getScm()).setReleaseTimestamp(plugin.getReleaseTimestamp()))
+            .map(pluginRepository::save)
+            .orElseGet(() -> pluginRepository.save(plugin));
     }
 }
--- a/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
@@ -24,6 +24,8 @@
 
 package io.jenkins.pluginhealth.scoring.service;
 
+import java.util.Optional;
+
 import io.jenkins.pluginhealth.scoring.model.Plugin;
 import io.jenkins.pluginhealth.scoring.repository.PluginRepository;
 
@@ -38,9 +40,19 @@ public class PluginService {
     }
 
     public Plugin saveOrUpdate(Plugin plugin) {
-        return pluginRepository.findByName(plugin.getName())
-            .map(pluginFromDatabase -> pluginFromDatabase.setScm(plugin.getScm()).setReleaseTimestamp(plugin.getReleaseTimestamp()))
-            .map(pluginRepository::save)
-            .orElse(pluginRepository.save(plugin));
+//        return pluginRepository.findByName(plugin.getName())
+//            .map(pluginFromDatabase -> pluginFromDatabase.setScm(plugin.getScm()).setReleaseTimestamp(plugin.getReleaseTimestamp()))
+//            .map(pluginRepository::save)
+//            .orElse(pluginRepository.save(plugin));
+        if (pluginRepository.findByName(plugin.getName()).isPresent()) {
+            Optional<Plugin> pluginFromDB = pluginRepository.findByName(plugin.getName());
+            pluginFromDB.get().setScm(plugin.getScm()).setReleaseTimestamp(plugin.getReleaseTimestamp());
+            pluginRepository.save(pluginFromDB.get());
+            return pluginFromDB.get();
+        }
+        else {
+            pluginRepository.save(plugin);
+            return plugin;
+        }
     }
 }
--- a/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
@@ -24,6 +24,8 @@
 
 package io.jenkins.pluginhealth.scoring.service;
 
+import java.util.Optional;
+
 import io.jenkins.pluginhealth.scoring.model.Plugin;
 import io.jenkins.pluginhealth.scoring.repository.PluginRepository;
 
@@ -35,6 +37,24 @@ public class PluginService {
 
     public PluginService(PluginRepository pluginRepository) {
         this.pluginRepository = pluginRepository;
+    }
+
+    public void updateDatabase(Plugin plugin) {
+        Optional<Plugin> pluginFromDatabase = pluginRepository.findByName(plugin.getName());
+
+        if (pluginFromDatabase.isPresent()) {
+            if (pluginFromDatabase.get().getReleaseTimestamp() != plugin.getReleaseTimestamp()) {
+                pluginFromDatabase.get().setReleaseTimestamp(plugin.getReleaseTimestamp());
+                saveOrUpdate(pluginFromDatabase.get());
+            }
+            if (pluginFromDatabase.get().getScm() != null && !pluginFromDatabase.get().getScm().equals(plugin.getScm())) {
+                pluginFromDatabase.get().setScm(plugin.getScm());
+                saveOrUpdate(pluginFromDatabase.get());
+            }
+        }
+        else {
+            saveOrUpdate(plugin);
+        }
     }
 
     public void saveOrUpdate(Plugin plugin) {
--- a/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
@@ -40,17 +40,10 @@ public class PluginService {
     }
 
     public void saveOrUpdate(Plugin plugin) {
-        Optional<Plugin> pluginFromDatabase = pluginRepository.findByName(plugin.getName());
-
-        if (pluginFromDatabase.isPresent()) {
-            pluginFromDatabase.get().setReleaseTimestamp(plugin.getReleaseTimestamp());
-            pluginFromDatabase.get().setScm(plugin.getScm());
-
-            pluginRepository.save(pluginFromDatabase.get());
-        }
-        else {
-            pluginRepository.save(plugin);
-        }
+        pluginRepository.findByName(plugin.getName())
+            .map(pluginFromDatabase -> pluginFromDatabase.setScm(plugin.getScm()).setReleaseTimestamp(plugin.getReleaseTimestamp()))
+            .map(updatedPluginFromDatabase -> pluginRepository.save(updatedPluginFromDatabase))
+            .orElse(pluginRepository.save(plugin));
     }
 
 }
--- a/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
@@ -24,8 +24,6 @@
 
 package io.jenkins.pluginhealth.scoring.service;
 
-import java.util.Optional;
-
 import io.jenkins.pluginhealth.scoring.model.Plugin;
 import io.jenkins.pluginhealth.scoring.repository.PluginRepository;
 
--- a/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
@@ -37,11 +37,10 @@ public class PluginService {
         this.pluginRepository = pluginRepository;
     }
 
-    public void saveOrUpdate(Plugin plugin) {
-        pluginRepository.findByName(plugin.getName())
+    public Plugin saveOrUpdate(Plugin plugin) {
+        return pluginRepository.findByName(plugin.getName())
             .map(pluginFromDatabase -> pluginFromDatabase.setScm(plugin.getScm()).setReleaseTimestamp(plugin.getReleaseTimestamp()))
-            .map(updatedPluginFromDatabase -> pluginRepository.save(updatedPluginFromDatabase))
+            .map(pluginRepository::save)
             .orElse(pluginRepository.save(plugin));
     }
-
 }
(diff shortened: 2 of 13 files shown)


src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -37,6 +39,24 @@ public PluginService(PluginRepository pluginRepository) {
         this.pluginRepository = pluginRepository;
     }
 
+    public void updateDatabase(Plugin plugin) {
+        Optional<Plugin> pluginFromDatabase = pluginRepository.findByName(plugin.getName());
+
+        if (pluginFromDatabase.isPresent()) {
+            if (pluginFromDatabase.get().getReleaseTimestamp() != plugin.getReleaseTimestamp()) {
```
dheerajodha, line 46: This condition is always true because `releaseTimestamp` of the plugin fetched from the DB has the zone ID of `Asia/Kolkata` and the plugin fetched from update-center.json file has zone ID of `UTC`.

I tried to include these changes one-by-one in the application.properties file because I want JPA to store timestamps with `UTC` zoneID:

* appended jdbc URL with: `?useLegacyDatetimeCode=false&serverTimezone=UTC`
* spring.jpa.properties.hibernate.jdbc.time_zone=UTC
* spring.jackson.time-zone=UTC

But I still see that the zone ID of the fetched plugins is not `UTC`. But the `releaseTimestamp` shows in UTC format when I run a select-query on the DB.

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -37,6 +39,24 @@ public PluginService(PluginRepository pluginRepository) {
...
+                pluginFromDatabase.get().setReleaseTimestamp(plugin.getReleaseTimestamp());
+                saveOrUpdate(pluginFromDatabase.get());
+            }
+            if (pluginFromDatabase.get().getScm() != null && !pluginFromDatabase.get().getScm().equals(plugin.getScm())) {
```
dheerajodha, line 50: Is this a suggested (or scalable) way to check for differences in values for a field? How can I make it look smarter and have less LOC? Can you please give some hints?

dheerajodha (commented):
I have a few questions, thanks for your time.

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
alecharp: I don't think we need to check the previous values here. The update-center will always have the most up-to-date values for SCM and releaseTimestamp.

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -37,6 +39,24 @@ public PluginService(PluginRepository pluginRepository) {
         this.pluginRepository = pluginRepository;
     }
 
+    public void updateDatabase(Plugin plugin) {
+        Optional<Plugin> pluginFromDatabase = pluginRepository.findByName(plugin.getName());
```
alecharp, line 43: Using `Optional` is a good idea but you are not using it to its potential. Look at the `map` and `orElse` method. You could reduce the number of lines using those, without compromising the readability of the code.

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
dheerajodha: So that would mean we'd run `saveOrUpdate()` method for every plugin. Meaning, that no matter whether the field values of a plugin record are the latest or not we would still be running an update query on the plugin records stored in the DB.

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
dheerajodha: Sure, I'm on it now

dheerajodha: [Correct me if I'm wrong]: As far as I could understand and based on the currently pushed code, the end result should be something like:
`pluginFromDatabase.map()`
`            .orElse(pluginRepository.save(plugin));`

But I'm not able to understand how to use the `map()` method. Can you please help?

alecharp: we could do something like

```java
pluginRepository.findByName(plugin.getName())
  .map(store -> store.withSCM(plugin.getSCM()).withReleaseTimestamp(plugin.getTimestamp()))
  .map(updated -> pluginRepository.save(updated)) // this can be a changed by a reference call
  .orElse(pluginRepository.save(plugin));
```

dheerajodha: Is this `withSCM()` and `withReleaseTimestamp` related to this: https://projectlombok.org/features/With?

alecharp: No, please don't add Lombok on the project. We don't need this. 
It's just the setters, in the "builder" format, so they return the instance. This allow us to link the calls.

dheerajodha: [Correct me if I'm wrong]: I read [this article](https://www.geeksforgeeks.org/builder-pattern-in-java/) and I modified the setter methods to return a `Plugin` instance so that we can chain multiple setter methods to eventually return a `Plugin` object instance.

dheerajodha:
@alecharp I'll add the tests as part of this PR then

dheerajodha:
@alecharp I have a question, where should the new tests be added?
Inside this class: https://github.com/jenkins-infra/plugin-health-scoring/blob/main/src/test/java/io/jenkins/pluginhealth/scoring/PluginHealthScoringIT.java?

Or inside a separate file called `PluginDuplicationTest.java`?

I can see a need to have a Postgres DB connected with the Spring project before I write tests for any methods. Because methods wouldn't return anything if the connections are not in place. So, I have a feeling I will reuse some part of what you wrote in the `PluginHealthScoringIT.java`. Any hints?

alecharp:
Technically, we could put all our tests in one Java class. But this is not what we should do.

Normally, we create a test class per class we want to test (`PluginService` -> `PluginServiceTest`), but we can also differentiate unit test (`PluginServiceTest`) and integration tests (`PluginServiceIT`). This is the general case, but you can also have technical limitations in which case you need to have multiple classes to cover the code a one class (because you need to load different configuration or something like that, or different database to connect to)

> I can see a need to have a Postgres DB connected with the Spring project before I write tests for any methods

If you want to have a real DB connected, then it is called an integration test. This is longer to setup (need to look into `testcontainers` project to setup a `postgresql` container before starting the test. https://www.testcontainers.org/quickstart/junit_5_quickstart/

If you want, you can _mock_ the repository layer, faking its interactions with the DB, in which case, we can say it's a unit test. But this also means you have to think more about what to fake, how to fake it etc. For example, the DB constrains (unique ID, unique name, etc.) would be handle directly by the real DB, you might need to fake that in a unit test. From what I used in the past, we can use Mockito to fake calls to methods and beans.

Here, I let you decide what test seems to be better, what configuration you need but please respect the conventions (naming, placement of the class, etc.)

Generally speaking it's good to have a test first, so that you can see it fails until you work on the code to fix the problem. But it's not always this way ;)

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import static org.mockito.Mockito.mock;
+
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@ContextConfiguration(initializers = PluginServiceTest.DockerPostgresDatasourceInitializer.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@Testcontainers
+@RunWith(MockitoJUnitRunner.class)
+public class PluginServiceTest {
```
dheerajodha, line 30: I chose to go with the Integration test rather than the unit test because I had trouble finding "what" and "how" to mock the pluginRepository and pluginService methods.

Then I tried to run this test class and I'm getting this error:
`org.junit.runners.model.InvalidTestClassError: Invalid test class 'io.jenkins.pluginhealth.scoring.PluginServiceTest':`
`  1. Test class should have exactly one public zero-argument constructor`

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
```diff
@@ -0,0 +1,65 @@
...
+            ).applyTo(applicationContext);
+        }
+    }
+
+    @Test
+    public void myTest() {
+        pluginService.saveOrUpdate(plugin);
+        pluginService.saveOrUpdate(plugin);
+
+        assertEquals(1, pluginRepository.findAllByName("myPlugin").size());
```
dheerajodha, line 63: I created a new method in Repository that fetches all the plugins from the DB with a given name. And here I'm fetching all the plugins by the name "myPlugin" and ensuring that there's only 1 (even though we saved the same plugin object twice in the database, see lines 60 and 61 above).

dheerajodha (commented):
I tried to work on the test last night as well, but got some errors. And since this is reducing our speed, I just thought to commit what I have for now and ask for hints from you instead.

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
dheerajodha, line 30: I chose to go with the Integration test rather than the unit test because I had trouble finding "what" and "how" to mock the pluginRepository and pluginService methods.

Then I tried to run this test class and I'm getting this error:
`org.junit.runners.model.InvalidTestClassError: Invalid test class 'io.jenkins.pluginhealth.scoring.PluginServiceTest':`
`  1. Test class should have exactly one public zero-argument constructor`

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
dheerajodha, line 63: I created a new method in Repository that fetches all the plugins from the DB with a given name. And here I'm fetching all the plugins by the name "myPlugin" and ensuring that there's only 1 (even though we saved the same plugin object twice in the database, see lines 60 and 61 above).

dheerajodha (commented):
I've committed my test file for your reference.
The test doesn't work, it throws an error. I was trying to solve it since last night but now I realized that this is slowing down our speed so with thick skin I committed the test.

Sorry, but I'd need a little help from you with this test. Can you please give some hints?

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
alecharp: Integration tests should be named `PluginServiceIT`.

alecharp: And please, put the test in the same package as the class it is validating.

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@ContextConfiguration(initializers = PluginServiceTest.DockerPostgresDatasourceInitializer.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@Testcontainers
+@RunWith(MockitoJUnitRunner.class)
```
alecharp, line 27: You are not using Mockito here, you can remove this.

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+public class PluginServiceTest {
+    private final PluginRepository pluginRepository;
+    private final PluginService pluginService;
+
+    Plugin plugin = new Plugin("myPlugin", "https://github.com/jenkinsci/my-plugin", null);
```
alecharp, line 32: You shouldn't create the plugin as a class field. We could want to test that we can create multiple plugins. 
Those kind of object should be limited to a test method.

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+            TestPropertyValues.of(
+                "spring.datasource.url=" + POSTGRE_SQL_CONTAINER.getJdbcUrl(),
+                "spring.datasource.username=" + POSTGRE_SQL_CONTAINER.getUsername(),
+                "spring.datasource.password=" + POSTGRE_SQL_CONTAINER.getPassword()
+            ).applyTo(applicationContext);
+        }
+    }
+
+    @Test
+    public void myTest() {
```
alecharp, line 57: I know naming is hard, but what about `shouldNotDuplicatePlugin` rather than `mytest`? This way, if the test ever fails, we know what we are testing without looking at its code.

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@ContextConfiguration(initializers = PluginServiceTest.DockerPostgresDatasourceInitializer.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@Testcontainers
+@RunWith(MockitoJUnitRunner.class)
+public class PluginServiceTest {
```
alecharp, line 28: Have you tried to run the test against the initial implementation of the `saveOrUpdate` method?

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
alecharp: You get this error in IntelliJ, you don't in `mvn verify`. I'm looking into that.

alecharp: this is because the runner of tests cannot use a non-zero test class constructor. It is ok to use non-final field with `@Autowired` annotation on the service and repository field in the test class.

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
dheerajodha: Noted. Thank you!

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
dheerajodha: > I don't think we should have multiple plugin with the same name.

Thanks, I've added a `unique=true` constraint on the name field. But regarding this quoted text: yes, we should not have multiple plugins with the same name but the aim of this test is to catch that test case where we mistakenly added two plugin objects with the same name in the database. And we're catching that by checking the size of the list returned by the `findAll()` method and equating it with "1". Whenever you're free, can you please help me understand this?

src/test/java/io/jenkins/pluginhealth/scoring/service/PluginServiceIT.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Test
+    public void shouldNotDuplicatePlugin() {
+        Plugin plugin = new Plugin("myPlugin", "https://github.com/jenkinsci/my-plugin", null);
+
+        pluginService.saveOrUpdate(plugin);
+        pluginService.saveOrUpdate(plugin);
+
+        assertEquals(1, pluginRepository.findAll().size());
```
dheerajodha, line 55: This test fails at this assertion with the error: `expected:<1> but was:<1878>`

Which is a little weird for me because I just saved one plugin object in the DB but somehow the test runs the full spring project (fetches all the data from the Update center and saves it into the DB) because only that's why there are 1877 (+1 object that we are adding as part of this test) plugins in the DB. Can you help me understand what am I missing?

alecharp: it is not weird. The initial import of the plugins is done on the main class of the application, which is ran when the test starts.

alecharp: I fixed that in https://github.com/jenkins-infra/plugin-health-scoring/pull/17/commits/f187fe3ca06b337b07e5fc42e02b9e0a6e935c63

src/test/java/io/jenkins/pluginhealth/scoring/PluginServiceTest.java
alecharp: I added a test in `PluginServiceIT`, which is failing (on purpose) for you to see my point. See https://github.com/jenkins-infra/plugin-health-scoring/pull/17/commits/4474bef52444fb5290b5311ea6ad7206a538386c.

There is some mechanism in hibernate which saving twice the same object is triggering, so your test is not entirely validating the issue we have. 
For you to understand, please see that the problem we have is triggered when we start the application a second time, with the database already setup (and populated). The second time, when we read the `update-center.json`, new objects are created based on the content of the file. The values of the objects might be the same, but the objects are different (different addresses in memory). This is not what your test was trying to reproduce. 

It's important, when fixing a bug, to understand the context in which the bug is reproducible, and stick to that context / execution when writing the test(s), so that we are truly reproducing the problem. And therefore, solving it.

alecharp: This is about learning. I'm thrilled that you have!

src/main/java/io/jenkins/pluginhealth/scoring/config/DatabaseConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.boot.autoconfigure.domain.EntityScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+@Configuration
+@EnableTransactionManagement
+@EnableJpaRepositories(basePackages = "io.jenkins.pluginhealth.scoring.repository")
+@EntityScan(basePackages = "io.jenkins.pluginhealth.scoring.model")
+public class DatabaseConfiguration {
```
dheerajodha, line 36: What is the need for this class?
I can see that you're telling, with the help of annotations, where you can find the repository and entity.
But where/when is it used? What is it replacing?

src/test/java/io/jenkins/pluginhealth/scoring/service/PluginServiceIT.java
```diff
@@ -0,0 +1,51 @@
...
+    public void shouldNotDuplicatePlugin() {
+        Plugin plugin = new Plugin("myPlugin", "https://github.com/jenkinsci/my-plugin", null);
+
+        pluginService.saveOrUpdate(plugin);
+        pluginService.saveOrUpdate(plugin);
+
+        assertThat(pluginRepository.findAll())
+            .hasSize(1)
+            .contains(plugin);
+    }
```
alecharp, line 34: I think we can get rid of this test because it is not properly validating the code.

alecharp (approved):
This looks good.

src/main/java/io/jenkins/pluginhealth/scoring/config/DatabaseConfiguration.java
alecharp: So that spring-boot is not scanning all the classes in the classpath for entities not for repositories.

alecharp (changes requested):
Two pieces of code to remove and we merge.

alecharp (approved):
Thanks

PR #18, opened 2022-07-11
Adding feature to store JSON data in DB

### LINKED ISSUE #16: Data storage of plugin details
When we will process the probes / rules, we will have to store a number of details on each plugin. 
In PostgreSQL, it is possible to store JSON data in a specific field. This would allow us to have some flexibility about the data we gather for each plugin.

 - Create a new field in Plugin class to store a key/value based data structure
 - Configure the project to create that field in the database

As a first usage, we can see that in the update-center JSON file, we have a list of deprecated plugins. Please store this detail in the newly create field.

Some links:
 - JSON data type in PostgreSQL: https://www.postgresql.org/docs/current/datatype-json.html
 - Mapping the data type to Java code: https://vladmihalcea.com/java-map-json-jpa-hibernate

dheerajodha:
Closes #16 

This PR does the following changes:
* Introduces a new field in the `Plugin` class: [Map<String, String> details](https://github.com/dheerajodha/plugin-health-scoring/blob/1bdbba5aafc22e6aa486b1f6dc0ee9920f96d472/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java#L60)
* There's a custom method: `Plugin addDetails(String key, String value)` ([here](https://github.com/dheerajodha/plugin-health-scoring/blob/1bdbba5aafc22e6aa486b1f6dc0ee9920f96d472/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java#L109)) which allows us to add a new JSON object of type <String, String>,
* [As seen here](https://github.com/dheerajodha/plugin-health-scoring/blob/1bdbba5aafc22e6aa486b1f6dc0ee9920f96d472/src/main/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterService.java#L69), we're adding a new JSON object, example:
`{"deprecation
_reason": "https://github.com/jenkinsci/jenkins/pull/5320"}` for those plugins which have been deprecated and are present in the DB.
* This is done by creating a new `record` called [UpdateCenterDeprecations](https://github.com/dheerajodha/plugin-health-scoring/blob/1bdbba5aafc22e6aa486b1f6dc0ee9920f96d472/src/main/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterService.java#L57) which helps us to fetch the "deprecations" object from the update-center.json file.
* Attaching a screenshot of how this json object looks like in the database now (see Record 4 below):
image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,6 +130,11 @@
 			<version>1.17.2</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.vladmihalcea</groupId>
+			<artifactId>hibernate-types-52</artifactId>
+			<version>2.16.3</version>
+		</dependency>
 	</dependencies>
 
 	<repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,7 @@
 		</dependency>
 		<dependency>
 			<groupId>com.vladmihalcea</groupId>
-			<artifactId>hibernate-types-52</artifactId>
+			<artifactId>hibernate-types-55</artifactId>
 			<version>2.16.3</version>
 		</dependency>
 	</dependencies>
--- a/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
@@ -101,7 +101,7 @@ public class Plugin {
     }
 
     public Map<String, String> getDetails() {
-        return new HashMap<>(details);
+        return Map.copyOf(details);
     }
 
     public Plugin addDetails(String key, String value) {
--- a/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
@@ -25,6 +25,7 @@
 package io.jenkins.pluginhealth.scoring.model;
 
 import java.time.ZonedDateTime;
+import java.util.Map;
 import java.util.Objects;
 import javax.persistence.Column;
 import javax.persistence.Entity;
@@ -33,8 +34,13 @@ import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.Table;
 
+import com.vladmihalcea.hibernate.type.json.JsonType;
+import org.hibernate.annotations.Type;
+import org.hibernate.annotations.TypeDef;
+
 @Entity
 @Table(name = "plugins")
+@TypeDef(name = "json", typeClass = JsonType.class)
 public class Plugin {
     @Id
     @GeneratedValue(strategy = GenerationType.AUTO)
@@ -49,14 +55,19 @@ public class Plugin {
     @Column(name = "release_timestamp")
     private ZonedDateTime releaseTimestamp;
 
+    @Type(type = "json")
+    @Column(columnDefinition = "jsonb")
+    private Map<String, String> details;
+
     public Plugin() {
 
     }
 
-    public Plugin(String name, String scm, ZonedDateTime releaseTimestamp) {
+    public Plugin(String name, String scm, ZonedDateTime releaseTimestamp, Map<String, String> details) {
         this.name = name;
         this.releaseTimestamp = releaseTimestamp;
         this.scm = scm;
+        this.details = details;
     }
 
     public long getId() {
@@ -85,6 +96,19 @@ public class Plugin {
 
     public void setReleaseTimestamp(ZonedDateTime releaseTimestamp) {
         this.releaseTimestamp = releaseTimestamp;
+    }
+
+    public Map<String, String> getDetails() {
+        return details;
+    }
+
+    public void setDetails(Map<String, String> details) {
+        this.details = details;
+    }
+
+    public Plugin addDetails(String key, String value) {
+        details.put(key, value);
+        return this;
     }
 
     @Override
--- a/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
@@ -25,6 +25,7 @@
 package io.jenkins.pluginhealth.scoring.model;
 
 import java.time.ZonedDateTime;
+import java.util.HashMap;
 import java.util.Map;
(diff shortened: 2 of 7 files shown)


dheerajodha:
@alecharp I believe this is not what you have in your mind exactly, maybe because I'm not storing the JSON data in a nested format.

And as we advance, we'd be using the JSON object in the Database to store intelligent nested formats, which are missing now. I'd need your help in understanding again how we plan to create nested objects in the DB and accordingly, I can make changes to the code.

src/main/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterService.java
```diff
@@ -50,16 +51,26 @@ public UpdateCenterService(ObjectMapper objectMapper, @Value("${jenkins.update.c
...
+        record UpdateCenter(Map<String, UpdateCenterPlugin> plugins, Map<String, UpdateCenterDeprecations> deprecations) {
         }
 
         UpdateCenter updateCenter = objectMapper.readValue(new URL(updateCenterURL), UpdateCenter.class);
-        return updateCenter.plugins.values().stream()
+        List<Plugin> pluginList = updateCenter.plugins.values().stream()
             .map(UpdateCenterPlugin::toPlugin)
             .collect(Collectors.toList());
+
+        for (Plugin plugin : pluginList) {
```
dheerajodha, line 67: I didn't feel happy to do this, because here I'm traversing the big 1800+ objects all over again to now add the "deprecation_reason" JSON object in them (wherever applicable). I understand this is now a scalable way to do this as it will slow things down as we move forward. Any thoughts on this?

dheerajodha (commented):
Sharing some of my doubts.

src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
```diff
@@ -49,14 +55,19 @@ public class Plugin {
     @Column(name = "release_timestamp")
     private ZonedDateTime releaseTimestamp;
 
+    @Type(type = "json")
+    @Column(columnDefinition = "jsonb")
+    private Map<String, String> details;
```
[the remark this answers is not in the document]
alecharp: This is bit restrictive, but it should work for now. 
What would happen in the future if we decide to have `{ 'maven': { 'parent': 4.39 }, 'jenkins': 2.347 }` in the details as well?

src/main/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterService.java
alecharp: You are correct, you are iterating over the plugins again, which might not be as performant as it could be.
Maybe we could leverage the fact that we do that already once (in lines 60 to 62) and transform the plugins to hold this details before returning the list. Something like a map operation ;)

src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
```diff
@@ -49,14 +55,19 @@ public class Plugin {
...
     public Plugin() {
 
     }
 
-    public Plugin(String name, String scm, ZonedDateTime releaseTimestamp) {
+    public Plugin(String name, String scm, ZonedDateTime releaseTimestamp, Map<String, String> details) {
         this.name = name;
         this.releaseTimestamp = releaseTimestamp;
         this.scm = scm;
+        this.details = details;
```
alecharp, line 70: I wouldn't set the map in the constructor. You can probably mark the map as `final` and initialize it by default to an empty map.

src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
```diff
@@ -87,6 +98,19 @@ public void setReleaseTimestamp(ZonedDateTime releaseTimestamp) {
         this.releaseTimestamp = releaseTimestamp;
     }
 
+    public Map<String, String> getDetails() {
+        return details;
+    }
+
+    public void setDetails(Map<String, String> details) {
+        this.details = details;
+    }
```
alecharp, line 107: I wouldn't set the map entirely.

alecharp (changes requested):

pom.xml
```diff
@@ -130,6 +130,11 @@
 			<version>1.17.2</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.vladmihalcea</groupId>
+			<artifactId>hibernate-types-52</artifactId>
```
alecharp, line 135:
```suggestion
			<artifactId>hibernate-types-55</artifactId>
```

src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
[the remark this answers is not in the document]
dheerajodha: As discussed, we'll be updating the Map's type in the future to make it include nested values.

alecharp:
Checkstyle warnings will be resolved when https://github.com/jenkins-infra/plugin-health-scoring/pull/17/commits/a5a2f6b906a8316ff35b8760ffe6e04f8b317295 is merged and ported here.

alecharp:
Please attend to the conflicts so we can move this forward.

alecharp (approved):
Thanks @dheerajodha.

src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
```diff
@@ -89,6 +100,15 @@ public Plugin setReleaseTimestamp(ZonedDateTime releaseTimestamp) {
         return this;
     }
 
+    public Map<String, String> getDetails() {
+        return new HashMap<>(details);
```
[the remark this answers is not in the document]
dheerajodha: Quoting from [here](https://stackoverflow.com/a/65659020): Regarding `Map.copyOf()`,

`But this has a price: with a copy, you need two times the amount of memory for the map (roughly). For really large maps, this may cause memory shortages up to an OutOfMemoryError`

Is this going to be a problem for us in the future?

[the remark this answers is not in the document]
alecharp: we can deal with this when it happens

Issue #11, opened 2022-06-28
Gather details from GitHub for each plugins


alecharp:
From the `update-center.json` file, we have the SCM link for each plugins. Based on this, we want to gather more details on each plugins. 

However, each plugins might not have the same set of possible details. For example, we can get the parent POM version if the plugin is using Maven as a build tool (over simplification, but let's say it's true for now). Because of this, we cannot process all the plugins the same way. The base example would be: some plugins don't have a SCM link or a valid one, then we cannot reach the repository to find out the latest commit date.

So, the rules engine needs to be able to do a cascading processing. For this, we can get inspiration from the Intercepting Servlet Filter chain. 

We have some families of rules which are only applicable if a preset is valid:
 - we can only get details from the source code of the plugin if we can get the source code of the plugin.

In this situation, we have:
 - a `RuleEngine` which have a list of `RulePreset`s.
 - each `RulePreset` has a list of `Rule`s which requires a specific element of the project (a SCM link, a POM file, etc)
 - each `Rule` are injected into the `RulePreset` they require 
 - each `Rule` are responsible to update the plugin they are applied to with the assertion they are made for (`hasValidSCM`, `isMavenProject`, `hasJenkinsfile`, `parentPOMVersion`, etc).

The idea is to not apply any source code rule if the SCM link is invalid. Same, we don't want to apply any Maven rules if the plugin is not built with Maven.

jleon33:
As discussed in our call on June 28th we will need documentation as well.

alecharp:
With #19 and other contributions around the `ProbeEngine` and `Probe` classes, this has been addressed.

PR #19, opened 2022-07-20
Probe engine 

alecharp:
relates to #11 

In this PR, I propose a data structure to hold the probe results and a way to run the probe over the plugins.
This is in no way the definitive structure nor way to run the probes, but a first approach.


src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
```diff
@@ -0,0 +1,66 @@
...
+        final Probe probe = mock(Probe.class);
+        final ProbeEngine probeEngine = new ProbeEngine(List.of(probe), pluginService);
+        final ProbeResult expectedResult = ProbeResult.success("foo", "bar");
+
+        when(probe.doApply(plugin)).thenReturn(expectedResult);
+        when(pluginService.streamAll()).thenReturn(Stream.of(plugin));
+        probeEngine.run();
+
+        verify(probe).doApply(plugin);
+        verify(pluginService).saveOrUpdate(plugin);
```
dheerajodha: The `verify` method is used to check how many times a method of a given mock object has been invoked.
So, in this case, we are cross-checking something like: "Hey, when we called `probeEngine.run();` we had to call the methods `doApply` and `saveOrUpdate` once in the subsequent calls." Right?

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Test
+    public void shouldBeAbleToRunSimpleProbe() {
+        final Plugin plugin = new Plugin("foo", "bar", ZonedDateTime.now());
+        final Probe probe = mock(Probe.class);
+        final ProbeEngine probeEngine = new ProbeEngine(List.of(probe), pluginService);
+        final ProbeResult expectedResult = ProbeResult.success("foo", "bar");
+
+        when(probe.doApply(plugin)).thenReturn(expectedResult);
+        when(pluginService.streamAll()).thenReturn(Stream.of(plugin));
```
dheerajodha, line 62: [Pls correct me if I'm wrong]: This part of the test is where we mock the return values of some methods of our mock objects (`probe`, `plugin`, and `pluginService`). It is needed because they are mock objects, when we use them we aren't going to their java class file to check their available methods, we will rely on these "when()" methods to tell us what these mock objects have to offer.

How do you decide which objects to mock? Since earlier you didn't define `plugin` as a mock object.

src/main/java/io/jenkins/pluginhealth/scoring/PluginHealthScoring.java
```diff
@@ -24,10 +24,14 @@
...
 package io.jenkins.pluginhealth.scoring;
 
+import io.jenkins.pluginhealth.scoring.config.GithubConfiguration;
+
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
+@EnableConfigurationProperties(value = GithubConfiguration.class)
```
dheerajodha, line 34: Why do we need to use this annotation and mention the GithubConfiguration.class?

I read about this annotation and as per my understanding, it's helpful in quickly registering a @ConfigurationProperties annotated class in Spring. And GithubConfiguration class does have the @ConfigurationProperties annotation. Then I read about the @ConfigurationProperties annotation and found out that it helps us to easily read values from the yaml file that you've created. And since the yaml file has so many values and we are only concerned with the github.oauth that's why you've mentioned prefix="github" so that's easier to pin down which exact field we need to read from the yaml file.

Is this correct?

src/main/java/io/jenkins/pluginhealth/scoring/probes/Probe.java
```diff
@@ -0,0 +1,59 @@
...
+     * Perform the analyze on a plugin
+     *
+     * @param plugin the plugin on which the analyze is done
+     * @return a ProbeResult representing the result of the analyze
+     */
+    protected abstract ProbeResult doApply(Plugin plugin);
+
+    protected abstract String key();
+
+    protected boolean requiresRelease() {
```
dheerajodha, line 56: What is our aim for having this boolean field?

Is it to give us an option like: For a given plugin,
- if `requiresRelease` returns `true` that would mean that we won't run the probe on it again if the last time we ran the probe was "after" the last release date of this plugin.
- if it returns `false` that would mean that we don't care if the last time we ran the probe was before or after the release date of this plugin, we will run the probes every single time regardless.

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -0,0 +1,112 @@
...
+            }
+            return Optional.empty();
+        }
+        try {
+            final GitHub gitHub = this.getGitHub();
+            if (LOGGER.isTraceEnabled()) {
+                GHRateLimit rateLimit = gitHub.getRateLimit();
+                LOGGER.trace("GitHub rate: {}/{}, reset: {}", rateLimit.getRemaining(), rateLimit.getLimit(), rateLimit.getResetDate());
+            }
+            final GHRepository repo = gitHub.getRepository(matcher.group("repo"));
```
dheerajodha, line 99: What do we mean by "repo" here? What's the ideal value for it?
I can see that if it's null then the scm is invalid.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
```diff
@@ -0,0 +1,97 @@
...
+        probeEngine.run();
+
+        verify(probe).doApply(plugin);
+        verify(plugin, atMostOnce()).addDetails(expectedResult);
+        verify(pluginService, atMostOnce()).saveOrUpdate(plugin);
```
dheerajodha, line 67: The verify method is used to check how many times a method of a given mock object has been invoked.
So, in this case, we are cross-checking something like: "Hey, when we called probeEngine.run(); we had to call the methods `doApply`, addDetails` and `saveOrUpdate` at most once (once, in case of "doApply") in the subsequent calls." Right?

In this way we're testing what happens when we call the `probeEngine.run()` method, we cannot do end-to-end testing so that's why we're relying on the amount of times a method is called to make sure everything works as expected.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
```diff
@@ -0,0 +1,97 @@
...
+            .addDetails(ProbeResult.success("wiz", "This is good"));
+        final Probe probe = mock(Probe.class);
+        final ProbeEngine probeEngine = new ProbeEngine(List.of(probe), pluginService);
+
+        when(probe.requiresRelease()).thenReturn(Boolean.TRUE);
+        when(probe.key()).thenReturn("wiz");
+        when(pluginService.streamAll()).thenReturn(Stream.of(plugin));
+        probeEngine.run();
+
+        verify(probe, never()).doApply(plugin);
```
dheerajodha, line 82: [Correct me if I'm wrong]: This is the part where we're testing when the requiresRelease returns true and the plugin's releaseTimestamp falls before the probeResult's timestamp.
In that case, we don't apply the probe to the plugin, we just simply return the previously calculated `probeResult`. Hence, we never had to call `probe.apply(plugin);`

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
```diff
@@ -0,0 +1,97 @@
...
+    public void shouldNotSaveErrors() throws Exception {
+        final Plugin plugin = mock(Plugin.class);
+        final Probe probe = mock(Probe.class);
+        final ProbeEngine probeEngine = new ProbeEngine(List.of(probe), pluginService);
+
+        when(probe.doApply(any(Plugin.class))).thenReturn(ProbeResult.error("foo", "bar"));
+        when(pluginService.streamAll()).thenReturn(Stream.of(plugin));
+        probeEngine.run();
+
+        verify(plugin, never()).addDetails(any());
```
dheerajodha, line 95: This test is to make sure that when our probe results in an error, we don't save it inside the plugin's `details` map (by calling the `addDetails` method).

Because you've filtered out the `ResultStatus.ERROR` status from reaching to `addDetails` method.

src/test/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbeTest.java
```diff
@@ -0,0 +1,73 @@
...
+    @Test
+    public void shouldRequireRelease() {
+        assertThat(new SCMLinkValidationProbe(new GithubConfiguration("")).requiresRelease()).isTrue();
+    }
+
+    @Test
+    public void shouldNotAcceptNullNorEmptyScm() {
+        final Plugin p1 = new Plugin("foo", null, ZonedDateTime.now());
+        final Plugin p2 = new Plugin("bar", "", ZonedDateTime.now());
+        final SCMLinkValidationProbe probe = new SCMLinkValidationProbe(new GithubConfiguration(""));
```
dheerajodha, line 53: Here, you're using an empty `oauth` String, why?
[Correct me if I'm wrong]: I guess, maybe because the following tests are just checking the scm URL's validity using the regex expression which doesn't require us to contact the server. Would we need to write any test which would require us to use a valid `oauth`? If yes, can you give one example?

dheerajodha (approved):
This review contains some questions and some statements which are there for me to cross-check my understanding from you. Thanks!

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -0,0 +1,121 @@
...
+    private Optional<GHRepository> fromSCMLink(@NotNull String scm) {
+        Matcher matcher = GH_PATTERN.matcher(scm);
+        if (!matcher.find()) {
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("{} is not respecting the SCM URL Template", scm);
+            }
+            return Optional.empty();
+        }
+        try {
+            final GitHub gitHub = this.getGitHub();
```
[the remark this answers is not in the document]
alecharp: `GitHub` is coming from the `github-api` library, which provides some useful methods to interact with GitHub API. In the current case, we only "validate" the link, to make sure the URL is correct. But we would use it to have more details, like the pull requests count, the open issues count, etc.

src/main/java/io/jenkins/pluginhealth/scoring/PluginHealthScoring.java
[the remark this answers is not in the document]
alecharp: This is a first step to have a configuration when we don't use `@Value` but a `ConfigurationBean.getJenkinsURL` for example. To have a more structured configuration file.

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -0,0 +1,121 @@
...
+            if (LOGGER.isTraceEnabled()) {
+                GHRateLimit rateLimit = gitHub.getRateLimit();
+                LOGGER.trace("GitHub rate: {}/{}, reset: {}", rateLimit.getRemaining(), rateLimit.getLimit(), rateLimit.getResetDate());
+            }
+            final GHOrganization org = gitHub.getOrganization(matcher.group("org"));
+            if (!"jenkinsci".equals(org.getLogin())) {
```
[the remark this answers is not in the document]
alecharp: As per https://www.jenkins.io/doc/developer/publishing/requesting-hosting/, the plugins should be hosted in `jenkinsci` organization. I know some are not (the SCM link is pointing to `jenkinsci` but there is a redirect). So I also wanted to mark such plugins are failing this probe.

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -0,0 +1,121 @@
...
+        if (!matcher.find()) {
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("{} is not respecting the SCM URL Template", scm);
+            }
+            return Optional.empty();
+        }
+        try {
+            final GitHub gitHub = this.getGitHub();
+            if (LOGGER.isTraceEnabled()) {
+                GHRateLimit rateLimit = gitHub.getRateLimit();
```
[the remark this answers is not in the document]
alecharp: This is correct. But I prevented its usage is normal situation and only call the rate limit API when running the application in the finer log level.

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -0,0 +1,121 @@
...
+                // TODO
+            }
+            return Optional.of(repo);
+        } catch (IOException ex) {
+            return Optional.empty();
+        }
+    }
+
+    private GitHub getGitHub() throws IOException {
+        return GitHub.connectUsingOAuth(githubConfiguration.getGitAccessToken());
```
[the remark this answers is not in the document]
alecharp: Any requests to the GitHub API requires the token. If not, then the requests are unauthenticated and you are only eligible to make 60 requests per hour, when authenticated requests are limited to 5000 per hour.

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -37,10 +40,16 @@ public PluginService(PluginRepository pluginRepository) {
         this.pluginRepository = pluginRepository;
     }
 
+    @Transactional
```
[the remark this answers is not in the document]
alecharp: This is correct. For example, if at some point we try to save the same plugin twice, but it has different details. This could lead to some issues.

src/main/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterService.java
```diff
@@ -58,18 +61,21 @@ Plugin toPlugin() {
...
         UpdateCenter updateCenter = objectMapper.readValue(new URL(updateCenterURL), UpdateCenter.class);
         return updateCenter.plugins().values().stream()
             .map(UpdateCenterPlugin::toPlugin)
             .map(plugin -> {
                 if (updateCenter.deprecations().containsKey(plugin.getName())) {
-                    return plugin.addDetails("deprecation_reason", updateCenter.deprecations().get(plugin.getName()).url());
-                }
-                else
+                    return plugin.addDetails(
+                        ProbeResult.failure(DEPRECATION_KEY, updateCenter.deprecations().get(plugin.getName()).url())
```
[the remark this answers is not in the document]
alecharp: That I don't know because, to be able to work correctly, this probe would need to have the update-center content.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
alecharp: correct.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
alecharp: my _personal_ thinking process on this is : Am I testing this class? How tedious is it to do a dummy implementation of this class? 

If you are not testing this class, probably it could be a mock.
If the class requires resources (DB access, etc) then it could be a mock.

src/main/java/io/jenkins/pluginhealth/scoring/probes/Probe.java
alecharp: Here, I added this field when testing the `SCMLinkValidationProbe`. The link can only be updated when a new release of the plugin is done (because only a new release can update the values in the update-center). And, at the same time, the probe was costly (API rate limitation). So, my reasoning was: when can I prevent in a clever way, the probe from being executed?
The response is: no need to execute the probe if the probe was previously executed on the plugin, and there is no new release of the plugin since the last execution.

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
alecharp: The `repo` group is coming from the `GH_REGEXP`

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
alecharp: exactly my thinking. I don't want to code a dummy probe and see that it provided a result. What I want is to see that the probe engine called the method `doApply` on the probe with the provided plugin, that it added the `ProbeResult` to the plugin and that we saved the plugin.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
alecharp: exactly.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
alecharp: exactly. I still don't know if it's a good idea or not. But for the moment I keep that. If I change my mind, I'll change this test as well.

This good because if we have some code where we rely on this behavior and we change it, we will see the problem at build time (on CI) and not when we run the code (production).

src/main/java/io/jenkins/pluginhealth/scoring/probes/Probe.java
dheerajodha: That's nice, makes sense.

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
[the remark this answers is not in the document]
dheerajodha: So, the current implementation doesn't check whether there's a redirect or not, right?
It just checks whether the "repo" field in the regex returns the "jenkinsci" or not.

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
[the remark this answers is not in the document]
dheerajodha: So even with 5000 reqs per hour, we are seeing a rate limit issue?

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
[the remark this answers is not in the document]
alecharp: Yes. Technically, I could get rid of the `github-api` dependency because of this.

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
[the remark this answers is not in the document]
alecharp: yes, because some requests are hidden. First I was doing "GitHub, given me this organization" and then "from this organization, give me the repository". This is at least 2 calls to GitHub. As we have 1883 plugins, it's 3766 requests. So yes, we (I) have to be careful about this.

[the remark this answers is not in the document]
alecharp: it's even worse than that: if you do a `curl -I https://api.github.com/repos/jenkinsci/mailer-plugin` (as in https://docs.github.com/en/rest/overview/resources-in-the-rest-api#detailed-representations), you get

```
[...]
x-ratelimit-limit: 60
x-ratelimit-remaining: 58
x-ratelimit-reset: ******
x-ratelimit-resource: core
x-ratelimit-used: 2
[...]
```

so it consumed 2 requests even though you only made one. Note the `60` is because I wasn't authenticated doing this request.

PR #20, opened 2022-07-25
Fetch Update Center info on a schedule

### LINKED ISSUE #9: Update Center data should be read once a day
The current implementation read the `update-center` when the application starts. 
However, we want, in the end, the application to run all the time and still read the update-center daily.

We need change the current behavior so that we can read the data without restarting the application.

- [ ] The delay between updates should be configurable
- [ ] The updates shouldn't require the application to restart

dheerajodha:
Closes #9

### DIFF
--- a/src/main/java/io/jenkins/pluginhealth/scoring/PluginHealthScoring.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/PluginHealthScoring.java
@@ -26,8 +26,10 @@ package io.jenkins.pluginhealth.scoring;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.scheduling.annotation.EnableScheduling;
 
 @SpringBootApplication(scanBasePackages = "io.jenkins.pluginhealth.scoring")
+@EnableScheduling
 public class PluginHealthScoring {
     public static void main(String[] args) {
         SpringApplication.run(PluginHealthScoring.class, args);
--- a/src/main/java/io/jenkins/pluginhealth/scoring/PluginHealthScoring.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/PluginHealthScoring.java
@@ -24,30 +24,14 @@
 
 package io.jenkins.pluginhealth.scoring;
 
-import java.io.IOException;
-
-import io.jenkins.pluginhealth.scoring.cli.ImportPluginCLR;
-
-import org.springframework.boot.CommandLineRunner;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
 @SpringBootApplication(scanBasePackages = "io.jenkins.pluginhealth.scoring")
 @EnableScheduling
-public class PluginHealthScoring implements CommandLineRunner {
-    private final ImportPluginCLR importPluginCLR;
-
-    public PluginHealthScoring(ImportPluginCLR importPluginCLR) {
-        this.importPluginCLR = importPluginCLR;
-    }
-
+public class PluginHealthScoring {
     public static void main(String[] args) {
         SpringApplication.run(PluginHealthScoring.class, args);
-    }
-
-    @Override
-    public void run(String... args) throws IOException {
-        importPluginCLR.updateDatabase();
     }
 }
--- a/src/main/java/io/jenkins/pluginhealth/scoring/PluginHealthScoring.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/PluginHealthScoring.java
@@ -24,14 +24,30 @@
 
 package io.jenkins.pluginhealth.scoring;
 
+import java.io.IOException;
+
+import io.jenkins.pluginhealth.scoring.cli.ImportPluginCLR;
+
+import org.springframework.boot.CommandLineRunner;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
 @SpringBootApplication(scanBasePackages = "io.jenkins.pluginhealth.scoring")
 @EnableScheduling
-public class PluginHealthScoring {
+public class PluginHealthScoring implements CommandLineRunner {
+    private final ImportPluginCLR importPluginCLR;
+
+    public PluginHealthScoring(ImportPluginCLR importPluginCLR) {
+        this.importPluginCLR = importPluginCLR;
+    }
+
     public static void main(String[] args) {
         SpringApplication.run(PluginHealthScoring.class, args);
+    }
+
+    @Override
+    public void run(String... args) throws IOException {
+        importPluginCLR.updateDatabase();
     }
 }
--- a/src/main/java/io/jenkins/pluginhealth/scoring/cli/ImportPluginCLR.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/cli/ImportPluginCLR.java
@@ -26,21 +26,30 @@ package io.jenkins.pluginhealth.scoring.cli;
 
 import java.io.IOException;
 
+import io.jenkins.pluginhealth.scoring.service.PluginService;
 import io.jenkins.pluginhealth.scoring.service.UpdateCenterService;
 
 import org.springframework.boot.CommandLineRunner;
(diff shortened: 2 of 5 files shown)


alecharp:
How do we configure this if we want it to run on thrusdays or if we want this to run every 15min?

src/main/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterService.java
```diff
@@ -48,6 +49,7 @@ public UpdateCenterService(@Value("${jenkins.update.center}") String updateCente
         this.updateCenterURL = updateCenterURL;
     }
 
+    @Scheduled(cron = "0 0 0 * * Mon", zone = "UTC")
```
dheerajodha, line 52: I think we'd just need to change this cron expression as per our need.

to run on Thursdays (at say, 2 am): "0 0 2 * * Thu"
to run every 15mins: "* */15 * * * *"

I'm referring this doc: https://spring.io/blog/2020/11/10/new-in-spring-5-3-improved-cron-expressions

alecharp: That's not configuration. The problem here is that if we want to change the execution schedule, we need to release the project.

dheerajodha: Do you mean we need to be able to input the "execution schedule" somewhere outside of the code so that we don't need to re-release the project every time just for this one change?

Would application.properties file be suitable to store the configuration?

lemeurherve: I'd suggest to use an environment variable to configure it.

dheerajodha: I've now shifted the cron string to the application.properties file and used that variable in the code.

alecharp:
This cannot be finished here. The `readUpdateCenter` method is not saving the plugins in the database. Fetch the `update-center.json` file on a schedule is not what we want.

src/main/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterService.java
```diff
@@ -36,18 +36,26 @@
...
 
-    public UpdateCenterService(@Value("${jenkins.update.center}") String updateCenterURL) {
+    public UpdateCenterService(@Value("${jenkins.update.center}") String updateCenterURL, PluginService pluginService) {
+        this.pluginService = pluginService;
         this.objectMapper = Jackson2ObjectMapperBuilder.json().build();
         this.updateCenterURL = updateCenterURL;
     }
 
+    @Scheduled(cron = "${cronexpression}", zone = "UTC")
+    public void updateDatabase() throws IOException {
```
dheerajodha, line 55: This is the newly introduced method and since I was using `pluginService` variable I had to initialise it in the constructor, which in turn told me about the problems in the UpdateCenterServiceTest.java file.

There I had to pass an object of `PluginService` class while creating an object of `UpdateCenterService` class.

src/test/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterServiceTest.java
```diff
@@ -30,22 +30,28 @@
...
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.stereotype.Service;
 
-@JsonTest
-@RunWith(SpringRunner.class)
-class UpdateCenterServiceTest {
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@DataJpaTest(includeFilters = @ComponentScan.Filter(Service.class))
+class UpdateCenterServiceTest extends AbstractDBContainerTest {
```
dheerajodha, line 46: As per my understanding, the tests were failing because now I'm using @Autowired annotation in this test so I need to build the application by extending this class to `AbstractDBContainerTest` and use `@DataJpaTest` to not load the entire application.

I removed the @JsonTest and @RunWith annotation because the tests pass without them, but I'm not sure this is the right decision so I wanted to confirm once.

src/main/java/io/jenkins/pluginhealth/scoring/cli/ImportPluginCLR.java
```diff
@@ -35,15 +34,13 @@
...
-    private final PluginService pluginService;
 
-    public ImportPluginCLR(UpdateCenterService updateCenterService, PluginService pluginService) {
+    public ImportPluginCLR(UpdateCenterService updateCenterService) {
         this.updateCenterService = updateCenterService;
-        this.pluginService = pluginService;
     }
 
     @Override
     public void run(String... args) throws IOException {
```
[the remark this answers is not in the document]
alecharp: Or you could transform the `ImportPluginCLR` class into a "normal" component, which takes a `PluginService` and a `UpdateCenterService` and has a scheduled method which does what the CLR does.

src/main/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterService.java
alecharp: I'm not sure it's write to have this in this service.

src/test/java/io/jenkins/pluginhealth/scoring/service/UpdateCenterServiceTest.java
alecharp: with https://github.com/jenkins-infra/plugin-health-scoring/pull/20#discussion_r930954167, we wouldn't have to make a DB connection to test the `UpdateCenterService`.

src/main/java/io/jenkins/pluginhealth/scoring/cli/ImportPluginCLR.java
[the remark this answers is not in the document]
dheerajodha: I'm trying to do it in 2 ways:
1. https://github.com/jenkins-infra/plugin-health-scoring/pull/20/commits/a19b3bf5dc81ab532def956ff3d5897ccb8dcffa
2. https://github.com/jenkins-infra/plugin-health-scoring/pull/20/commits/9130cf96fbdd0a5b17c8c9345419ba44a37ce773
    2.1 I'm getting an error here, so trying to solve it now.

Does any one of these approaches look similar to what you're expecting?

src/main/java/io/jenkins/pluginhealth/scoring/PluginHealthScoring.java
```diff
@@ -24,14 +24,30 @@
...
+
+import org.springframework.boot.CommandLineRunner;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
 @SpringBootApplication(scanBasePackages = "io.jenkins.pluginhealth.scoring")
 @EnableScheduling
-public class PluginHealthScoring {
+public class PluginHealthScoring implements CommandLineRunner {
```
alecharp, line 38: Definitively not that. We removed that because you have all your tests using the app that will run the code.

alecharp (changes requested):

src/main/java/io/jenkins/pluginhealth/scoring/cli/ImportPluginCLR.java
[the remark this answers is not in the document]
dheerajodha: I think I didn't understand what you meant by the "normal" component but I am now taking both `PluginService` and `UpdateCenterService` and have a scheduled method (called `updateDatabase()`) here https://github.com/jenkins-infra/plugin-health-scoring/pull/20/commits/169a464e0e7d6e80e82043f8d6d66b72233820a8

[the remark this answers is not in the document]
alecharp: My point is that we don't need a CommandLineRunner implementation anymore. We could have a bean (https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html) which has a scheduled method to update the database based on the update-center. 
This bean could be clever and know when the last update occurred and decide to update the database when it's loaded right away if the last update was done 2days ago.

alecharp (approved):

alecharp (approved):

PR #24, opened 2022-08-05
Using env variable to store cron expression

### LINKED ISSUE #21: Fix update-center read scheduling expression
The current CRON configuration cannot be modified without releasing a new version of the project. 
We need this expression to be set from environment variable so that we can modify it and only need to restart the application.

dheerajodha:
Closes #21


alecharp:
This is not working as you think it is. See the failing test.

dheerajodha:
ah I was just googling and googling why the placeholder value ${SCHEDULE_CRONEXPRESSION} is not getting read by the application. I thought it was the fault of how I've written the code.

But then it struck me that "hey, since this one test is failing, let's check its definition" and then I saw that I'm not providing the applicationContext with the variable "cronexpression" and that's why it is complaining that it doesn't have all the variable it needs to safely build the application (pls correct me if I'm wrong).

alecharp (approved):

PR #26, opened 2022-08-11
Enabling Batch Updates

dheerajodha:
This PR aims to:
- [ ] provide the feature of "batch updates" so that, from now on, we can decrease the time we spend interacting with the database.
- [x] fix the `ERROR: duplicate key value violates unique constraint` as seen while running the project.

### DIFF
--- a/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
@@ -26,14 +26,15 @@ package io.jenkins.pluginhealth.scoring.service;
 
 import java.util.List;
 import java.util.stream.Stream;
-import javax.transaction.Transactional;
 
 import io.jenkins.pluginhealth.scoring.model.Plugin;
 import io.jenkins.pluginhealth.scoring.repository.PluginRepository;
 
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 
 @Service
+@Transactional(readOnly = true)
 public class PluginService {
     private final PluginRepository pluginRepository;
 
@@ -49,11 +50,11 @@ public class PluginService {
             .orElseGet(() -> pluginRepository.save(plugin));
     }
 
-    @Transactional
     public Stream<Plugin> streamAll() {
         return pluginRepository.findAll().stream();
     }
 
+    @Transactional
     public List<Plugin> batchUpdate(List<Plugin> pluginList) {
         return pluginRepository.saveAll(pluginList.stream()
             .map(plugin -> pluginRepository.findByName(plugin.getName())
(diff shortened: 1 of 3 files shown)


src/main/resources/application.yaml
```diff
@@ -7,6 +7,9 @@ spring:
     jpa.properties.hibernate:
         dialect: org.hibernate.dialect.PostgreSQLDialect
         jdbc.lob.non_contextual_creation: true
+        jdbc.batch_size: ${BATCH_SIZE}
```
alecharp, line 10: Why would we want this configurable?

src/main/resources/application.yaml
```diff
@@ -7,6 +7,9 @@ spring:
...
+        order_inserts: true
+        order_updates: true
```
alecharp, line 12: What are those for?

src/main/resources/application.yaml
dheerajodha: While I was trying batch updates with different batch sizes, I noticed that the batch size needs to be picked very carefully.
As per my understanding, increasing the batch size wouldn't necessarily improve the performance and the same for decreasing the batch size. So, I thought the batch size cannot be constant because it really varies with the number of objects we have to update, and that is why I made it configurable.

But since the update center rarely adds or deletes a new plugin, I guess we can keep it constant if we like. What do you think?

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -24,15 +24,17 @@
...
-import javax.transaction.Transactional;
 
 import io.jenkins.pluginhealth.scoring.model.Plugin;
 import io.jenkins.pluginhealth.scoring.repository.PluginRepository;
 
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 
 @Service
+@Transactional(readOnly = true)
```
dheerajodha, line 37: I read this article: https://vladmihalcea.com/spring-transactional-annotation/ where the author suggests:

> For transactional services, it’s good practice to set the readOnly attribute to true at the class level and override it on a per-method basis for the service methods that need to write to the database.

Keeping this in mind, I removed the @Transactional annotation from the `streamAll` method since as per my understanding, it is a read-only transaction.


I'll undo the change if you find this change wrong and/or unnecessary.

dheerajodha (commented):
A change related to @Transactional annotation. Thoughts?

src/main/resources/application.yaml
alecharp: I'd prefer to have a fix value because here the application would fail to start without the variable.

dheerajodha:
I have noticed that currently, the batching works on insert statements but not on update statements.

dheerajodha:
Here's what I see while running the application on an empty database:
image

Here's what I see while running the application on the non-empty database:
image

alecharp:
Let's step away from this for now. We will revisit the problem when we see a problem of performance coming from the database.

alecharp:
Closing for now.

Issue #23, opened 2022-08-03
Highlight project improvement 


alecharp:
Due to the current state of the project, it is difficult for anyone outside the developers team to know what the project is doing, how it is evolving.

We need to find a way to show the list of probes currently implemented on the project. We could also show a page with the result of the execution of each probes on the plugins we know.

dheerajodha:
What do you mean by "page"? is it just a markdown file with a hardcoded list of current probes (which I can understand would not be scalable) or some page that takes values dynamically from our project (something like JSP? maybe...)?

alecharp:
I had a JSP in mind. However, if we could generate a static page which could be generated once and stored by the CI, it could be great as well.

alecharp:
With #27 and the `probe` label here on the GitHub repository, we have technical ways to see the current status as well as the short-term plan on this project concerning the probe implementation.

For a larger view of the plan, we would need a page listing more long-term plans but I don't see any way to maintain that correctly.

I'm considering this as resolved, but we might want to reconsider this later on.

PR #27, opened 2022-08-16
Listing probes

alecharp:
This is creating a webpage in `/probes` URI to list the available probes of the application. 

This is related to #23.


src/main/resources/templates/probes/listing.html
```diff
@@ -0,0 +1,24 @@
...
+<body>
+
+<main data-layout-fragment="content">
+    <div class="probes-listing--table">
+        <div class="probes-listing--header">
+            <span>ID</span>
+            <span>Name</span>
+            <span>Description</span>
+        </div>
+        <div data-th-each="probe : ${probes}" class="probes-listing--row">
```
dheerajodha, line 15: This is how we're using the object from java here in this HTML file. We're iterating over `${probes}` which has the same as our `ModelAndView` object.

src/main/resources/templates/layouts/main.html
```diff
@@ -0,0 +1,25 @@
...
+    <script src="http://localhost:35729/livereload.js"
+            data-th-if="${#arrays.contains(@environment.getActiveProfiles(), 'dev')}"></script>
+</head>
+<body>
+
+<div id="app">
+    <header>
+        <span class="app--title">Plugin Health Scoring</span>
+    </header>
+    <main data-layout-fragment="content">
```
dheerajodha, line 20: I was wondering how this main page is showing details of the listing.html file, I guess this `<main>` tag with attribute `data-layout-fragment="content"` is rendering values of the listing.html file here on this page because that tag contains the rendering logic of each probe.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbesController.java
```diff
@@ -0,0 +1,59 @@
...
+public class ProbesController {
+    private final List<Probe> probes;
+
+    public ProbesController(List<Probe> probes) {
+        this.probes = probes;
+    }
+
+    @GetMapping(path = "")
+    public ModelAndView list() {
+        final ModelAndView modelAndView = new ModelAndView("probes/listing");
```
dheerajodha, line 48: `ModelAndView`: This is so cool!
So this is how would we send specific objects from the backend to the frontend files.
Here, the `viewName` is "probes/listing", and its corresponding UI file is placed relative to the "templates" folder at `probes/listing.html`, that's how the file is located where these objects would be made accessible.

dheerajodha (approved):
Makes sense.
No direct questions. Just some "Correct me if I'm wrong" statements that can be ignored if they make sense.

src/main/resources/templates/layouts/main.html
alecharp: this is only because of the `data-layout-fragment` attribute. This is coming from https://ultraq.github.io/thymeleaf-layout-dialect/processors/decorate/. I could have only use `data-th-fragment` from plain Thymeleaf templating as well: https://www.thymeleaf.org/doc/tutorials/3.0/usingthymeleaf.html#template-layout.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbesController.java
alecharp: correct.

src/main/resources/templates/probes/listing.html
alecharp: it's from the model part of the `ModelAndView`, yes.

dheerajodha: Yes, we defined the name of the view in the constructor and added model data to `ModelAndView` via the `addObject()` method.

src/main/resources/templates/layouts/main.html
dheerajodha: I'm able to make more sense now after reading the first link you shared. This main.html layout acts as a sort of ‘default’, with our probe-listing content acting as implementations (which are defined inside of the fragment we've created) on top of this default.

PR #32, opened 2022-08-19
Creates a context across probe executions for a plugin

### LINKED ISSUE #31: Probe engine execution context
Some probes implementation will have a hierarchy. 
This means that, some probes will require the validation of others before being executed but also get the context the previous probes created. 

For example, the `SCMLinkValidationProbe` is useful to make sure the `scm` field of the plugin is correct and once it is validated, it can be used to clone said repository. The repository might be clone in a specific folder which should be passed to another probe. This way the repository is cloned once. 

This also means that probes might have some cleanup tasks after their execution.

alecharp:
Resolves #31 

This create an environment where one probe could use a previously executed probe result to condition its execution.

### DIFF
--- a/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngine.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngine.java
@@ -62,7 +62,7 @@ public class ProbeEngine {
             .peek(plugin -> {
                 probes.forEach(probe -> {
                     try {
-                        var previousResult = plugin.getDetails().get(probe.key());
+                        final ProbeResult previousResult = plugin.getDetails().get(probe.key());
                         if (previousResult == null ||
                             (probe.requiresRelease()
                                 && previousResult.timestamp() != null
--- a/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngine.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngine.java
@@ -25,7 +25,6 @@
 package io.jenkins.pluginhealth.scoring.probes;
 
 import java.util.List;
-import java.util.stream.Collectors;
 
 import io.jenkins.pluginhealth.scoring.model.ProbeResult;
 import io.jenkins.pluginhealth.scoring.model.ResultStatus;
@@ -59,33 +58,33 @@ public class ProbeEngine {
      */
     public void run() {
         LOGGER.info("Start running probes on all plugins");
-        if (LOGGER.isDebugEnabled()) {
-            LOGGER.debug("Registered probes: {}", probes.stream().map(Probe::key).collect(Collectors.joining(", ")));
-        }
         pluginService.streamAll().parallel()
-            .peek(plugin -> probes.stream()
-                .map(probe -> {
+            .peek(plugin -> {
+                probes.forEach(probe -> {
                     try {
-                        if (probe.requiresRelease() && plugin.getDetails().containsKey(probe.key())) {
-                            ProbeResult probeResult = plugin.getDetails().get(probe.key());
-                            if (probeResult.timestamp() != null && plugin.getReleaseTimestamp().isBefore(probeResult.timestamp())) {
-                                if (LOGGER.isDebugEnabled()) {
-                                    LOGGER.debug("{} requires a release of {} to process it again.", probe.key(), plugin.getName());
-                                }
-                                return probeResult;
+                        var previousResult = plugin.getDetails().get(probe.key());
+                        if (previousResult == null ||
+                            (probe.requiresRelease()
+                                && previousResult.timestamp() != null
+                                && previousResult.timestamp().isBefore(plugin.getReleaseTimestamp()))
+                        ) {
+                            if (LOGGER.isTraceEnabled()) {
+                                LOGGER.trace("Running {} on {}", probe.key(), plugin.getName());
+                            }
+                            final ProbeResult result = probe.apply(plugin);
+                            if (result.status() != ResultStatus.ERROR) {
+                                plugin.addDetails(result);
+                            }
+                        } else {
+                            if (LOGGER.isDebugEnabled()) {
+                                LOGGER.debug("{} requires a release of {} to process it again.", probe.key(), plugin.getName());
                             }
                         }
-                        if (LOGGER.isTraceEnabled()) {
-                            LOGGER.trace("Running {} on {}", probe.key(), plugin.getName());
-                        }
-                        return probe.apply(plugin);
                     } catch (Throwable t) {
                         LOGGER.error("Couldn't run {} on {}", probe.key(), plugin.getName(), t);
-                        return ProbeResult.error(probe.key(), "Could not run probe " + probe.key());
                     }
-                })
-                .filter(result -> result.status() != ResultStatus.ERROR)
-                .forEach(plugin::addDetails))
+                });
+            })
             .forEach(pluginService::saveOrUpdate);
         LOGGER.info("Probe engine has finished");
     }
--- a/src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
+++ b/src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
@@ -24,9 +24,12 @@
 
 package io.jenkins.pluginhealth.scoring.probes;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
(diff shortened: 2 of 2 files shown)


src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngine.java
```diff
@@ -59,33 +58,33 @@ public ProbeEngine(List<Probe> probes, PluginService pluginService) {
...
     public void run() {
         LOGGER.info("Start running probes on all plugins");
-        if (LOGGER.isDebugEnabled()) {
-            LOGGER.debug("Registered probes: {}", probes.stream().map(Probe::key).collect(Collectors.joining(", ")));
-        }
         pluginService.streamAll().parallel()
-            .peek(plugin -> probes.stream()
-                .map(probe -> {
+            .peek(plugin -> {
+                probes.forEach(probe -> {
```
dheerajodha, line 63: In main branch, currently, we're using `probes.stream().forEach()` and here you've replaced that with`probes.forEach()`.

The main aim of this PR is to pass the context of 1 probe to another probe in a pre-determined sequential way. And as per the check [done by me here](https://github.com/jenkins-infra/plugin-health-scoring/blob/172e52b3ebd8bc26301185f02f21cc1bfc59d4f4/src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java#L39), this won't work because it is using `probes.stream().forEach()` therefore there is a possibility that my new probe would run "before" the `SCMLinkValidationProbe`. I have questions:

1. While using `probes.stream().forEach()`, would our probes be running as per the `ORDER` defined by us in each probe file?
2. While using `probes.forEach()`, the probes would be running based on the iterator which uses the value of `ORDER` variable to order the execution of each probe. Is this correct?

alecharp: there is no guaranty of order of execution with the Stream.
In our case, because we might need a previous probe result to execution the current probe, we need to be sure of the execution order. So, we need to be sure that right after the probe is execution, that we have a result, we put this result into the plugin details.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
```diff
@@ -92,4 +95,40 @@ public void shouldNotSaveErrors() {
...
+    public void shouldBeAbleToGetPreviousContextResultInExecution() throws Exception {
+        final Plugin plugin = spy(Plugin.class);
+        final Probe probeOne = new Probe() {
+            @Override
+            protected ProbeResult doApply(Plugin plugin) {
+                return ProbeResult.success(key(), "This is ok");
+            }
+
+            @Override
+            protected String key() {
```
dheerajodha, line 109: Java Compiler is complaining about `ERROR: key() in cannot override key() in io.jenkins.pluginhealth.scoring.probes.Probe`. I'm guessing it's because you've used an Anonymous Subclass, am I right?

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
```diff
@@ -92,4 +95,40 @@ public void shouldNotSaveErrors() {
...
+                    ProbeResult.success(key(), "This is also ok") :
+                    ProbeResult.error(key(), "This cannot be validated");
+            }
+
+            @Override
+            protected String key() {
+                return "bar";
+            }
+        };
+        final ProbeEngine probeEngine = new ProbeEngine(List.of(probeOne, probeTwo), pluginService);
```
dheerajodha, line 126: In which order of the probes would `probes.forEach` process the 2 probes here?
Because the test would fail if probeTwo gets processed first, I was wondering about how the order is picked up.
Would the test fail if we pass `List.of(probeTwo, probeOne)`?

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
```diff
@@ -92,4 +95,40 @@ public void shouldNotSaveErrors() {
 
         verify(plugin, never()).addDetails(any());
     }
+
+    @Test
+    public void shouldBeAbleToGetPreviousContextResultInExecution() throws Exception {
+        final Plugin plugin = spy(Plugin.class);
```
dheerajodha, line 101: Are you using `spy()` on the `Plugin` class because then you won't have to stub the `addDetails()` (and all the other methods inside this class) method and just let it use its concrete implementation?

alecharp: yes.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
alecharp: no because it should be a `public` method now.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngineTest.java
alecharp: From the `List.of` method implementation, the elements are ordered as they are in parameters. Please look at it.

> Would the test fail if we pass List.of(probeTwo, probeOne)?

Yes. That is how I implemented the `probeTwo`. If it cannot get the probe result from the first probe, it returns an error which are not added into the plugin details. So, we wouldn't have 2 access to `addDetails` method and we wouldn't have  the keys `foo` and `bar` into the details map as key.

dheerajodha: Thanks for answering in-depth, this is helpful. I read about `List.of` (I should've googled it in the first place) and it helped me to understand this.

dheerajodha (approved):

src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngine.java
```diff
@@ -59,33 +58,33 @@ public ProbeEngine(List<Probe> probes, PluginService pluginService) {
...
+                            if (LOGGER.isTraceEnabled()) {
+                                LOGGER.trace("Running {} on {}", probe.key(), plugin.getName());
+                            }
+                            final ProbeResult result = probe.apply(plugin);
+                            if (result.status() != ResultStatus.ERROR) {
+                                plugin.addDetails(result);
+                            }
+                        } else {
+                            if (LOGGER.isDebugEnabled()) {
+                                LOGGER.debug("{} requires a release of {} to process it again.", probe.key(), plugin.getName());
```
dheerajodha, line 80: If I have a probe that doesn't require a release to happen for it to be run again, then how often would `probe.apply()` run?

For example, the LastCommitDate probe doesn't require a plugin release because the commits are going to be added regardless of the release (I think) so, that info can change from time to time and we are not dependent on the fact whether the plugin has a new release or not. Based on this then, `probe.apply(plugin);` would run only once, right?

alecharp: > If I have a probe that doesn't require a release to happen for it to be run again, then how often would probe.apply() run?

if the probe requires a release, and it was already ran on the plugin after the last release, then it won't be re-run. If you have a doubt, please add a test here.

dheerajodha: That's true, but I'm asking about the probes which do not require a release, and have been run once already. How often would `probe.apply()` run for that probe? Based on the code, I see that the probe won't run again. But our last-commit-date can change regardless of a plugin release, so I think it wouldn't get updated after the first time we run it.

dheerajodha:
I'm trying to push tests to your PR but getting a "permission denied" error. Can you check the box in the PR which says “Allow edits from maintainers.”?

alecharp:
> Can you check the box in the PR which says “Allow edits from maintainers.”? 

it's already activated.

src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngine.java
```diff
@@ -59,33 +58,33 @@ public ProbeEngine(List<Probe> probes, PluginService pluginService) {
...
-                        return ProbeResult.error(probe.key(), "Could not run probe " + probe.key());
                     }
-                })
-                .filter(result -> result.status() != ResultStatus.ERROR)
-                .forEach(plugin::addDetails))
+                });
+            })
             .forEach(pluginService::saveOrUpdate);
         LOGGER.info("Probe engine has finished");
     }
```
dheerajodha, line 90: I need to add more tests for this file, more specifically, for the logic inside this `run()` method.
I don't see any other use case to cover, how do you decide if there are any more unit tests that can be added?

For example, here's what I know we should write a test for:
1. `final ProbeResult previousResult = plugin.getDetails().get(probe.key());`
     We should test whether the previousResult of a probe is being accessible its next run, you've already written a test for it.
2. The `if` condition on line 66:
    * A test for `null` previousResult
    * A test for non-null previousResult
    * A test for a probe that requires a release, and another which doesn't
    * A test for a probe whose timestamp is before the relaseTimestamp of the plugin, and vice versa
3. `if` condition on line 75, which makes sure we don't call `addDetails` on a probe with `ResultStatus == ERROR`

And for all the above scenarios we have already written tests. Can you please suggest how can we further increase the test coverage?

PR #33, opened 2022-08-23
Probe: Gets the last commit date of the repository

### LINKED ISSUE #28: probe: Get the date of the last commit of the repository
It is interesting to know when the last commit on the repository was made to help determine the plugin maintenance health.

- [ ] Store in DB the last commit date.

dheerajodha:
Closes #28


src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
```diff
@@ -0,0 +1,78 @@
...
+@Component
+@Order(value = LastCommitDateProbe.ORDER)
+public class LastCommitDateProbe extends Probe {
+    private static final Logger LOGGER = LoggerFactory.getLogger(LastCommitDateProbe.class);
+
+    public static final int ORDER = SCMLinkValidationProbe.ORDER + 1;
+
+    @Override
+    public ProbeResult doApply(Plugin plugin) {
+        final String scmLinkValidationProbeKey = new SCMLinkValidationProbe(null, null).key();
```
alecharp, line 31: You shouldn't need to instantiate the probe only to get its `key`. If you need it, please refactor that probe to have access to the key.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
```diff
@@ -0,0 +1,78 @@
...
+
+        try {
+            if (plugin.getDetails().get(scmLinkValidationProbeKey) == null) {
+                LOGGER.error("Couldn't run {} on {} because previous SCMLinkValidationProbe has null value in database", key(), plugin.getName());
+                return ProbeResult.error(key(), "SCM link has not been probed yet");
```
alecharp, line 36: shouldn't be `LOGGER.error` but `LOGGER.warning` and not a `ProbeResult.error` but `ProbeResult.failure`. Errors are not saved in DB.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
```diff
@@ -0,0 +1,78 @@
...
+            }
+
+            if (plugin.getDetails().get(scmLinkValidationProbeKey).status().equals(ResultStatus.SUCCESS)) {
```
alecharp, line 39: `ResultStatus` is an `enum`, you can use `==`.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
```diff
@@ -0,0 +1,78 @@
...
+                File dir = new File(System.getProperty("java.io.tmpdir") + "/" + plugin.getName());
```
alecharp, line 40: what about `Files.createTempDir` (not sure about the name, to be checked)?

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
```diff
@@ -0,0 +1,78 @@
...
+
+                Git git = Git.cloneRepository()
+                    .setURI(plugin.getScm())
+                    .setDirectory(dir)
+                    .call();
```
alecharp, line 45: Is is possible to do a `shallow` clone, which would not bring all the history of the repository? I don't think we need all the history here, correct?

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
```diff
@@ -0,0 +1,78 @@
...
+                deleteDirectory(dir);
+                return ProbeResult.success(key(), String.valueOf(lastCommitDateTime));
+            }
+            else {
+                return ProbeResult.failure(key(), "Due to invalid SCM, latest commit date cannot be found");
+            }
+        }
+        catch (GitAPIException e) {
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Invalid SCM link caused {}", e.getMessage());
```
alecharp, line 59: Is it the only reason to get this exception? Because if we use the validation of the `SCMLinkValidationProbe`, we shouldn't have this.

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -44,6 +46,12 @@ public PluginService(PluginRepository pluginRepository) {
     public Plugin saveOrUpdate(Plugin plugin) {
         return pluginRepository.findByName(plugin.getName())
             .map(pluginFromDatabase -> pluginFromDatabase.setScm(plugin.getScm()).setReleaseTimestamp(plugin.getReleaseTimestamp()))
+            .map(pluginFromDatabase -> {
+                for (Map.Entry<String, ProbeResult> freshlyReadProbe : plugin.getDetails().entrySet()) {
+                    pluginFromDatabase.addDetails(freshlyReadProbe.getValue());
+                }
+                return pluginFromDatabase;
```
alecharp, line 53: Could we have a new method this would take a map and add its content to the object?

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
```diff
@@ -0,0 +1,61 @@
...
+import io.jenkins.pluginhealth.scoring.model.ResultStatus;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
```
alecharp, line 16: I don't see any mock, why do you need this?

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    public void shouldNotRequireRelease() {
+        assertThat(new LastCommitDateProbe().requiresRelease()).isFalse();
+    }
+
+    @Test
+    public void shouldReturnSuccessStatusOnValidSCM() {
+        final Plugin plugin = new Plugin("parameterized-trigger", "https://github.com/jenkinsci/parameterized-trigger-plugin.git", ZonedDateTime.now())
+            .addDetails(ProbeResult.success("scm", "The plugin SCM link is valid"));
```
alecharp, line 31: Does this means we are cloning an external repository? What happens if the repository change? If, during the tests execution, there is no connection to internet? Or that GitHub is down?

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+        final ProbeResult r = probe.apply(plugin);
+
+        assertThat(r.id()).isEqualTo("last-commit-date");
+        assertThat(r.status()).isEqualTo(ResultStatus.SUCCESS);
+    }
+
+    @Test
+    public void shouldReturnFailureOnInvalidSCM() throws GitAPIException {
+        final Plugin plugin = new Plugin("parameterized-trigger", "https://github.com/jenkinsci/random-repo.git", ZonedDateTime.now())
```
alecharp, line 42: What happens if someone creates a repository with this name in the future?

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Test
+    public void shouldFailToRunAsFirstProbe() {
+        final Plugin plugin = new Plugin("parameterized-trigger", "https://github.com/jenkinsci/random-repo.git", ZonedDateTime.now());
+        final LastCommitDateProbe probe = new LastCommitDateProbe();
+
+        final ProbeResult r = probe.apply(plugin);
+
+        assertThat(r.status()).isEqualTo(ResultStatus.ERROR);
```
alecharp, line 58: I understand the idea behind the test, but this shouldn't be the responsibility of the probe to ensure this, as a it's controlled with `@Order`. 
You could write the test testing that the order of this probe is larger that the probe it should be after.

alecharp (changes requested):

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
alecharp: it's not supported by JGit (https://bugs.eclipse.org/bugs/show_bug.cgi?id=475615)

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
```diff
@@ -0,0 +1,77 @@
...
+                    final Path tempDirectory = Files.createTempDirectory(plugin.getName());
+                    try (Git git = Git.cloneRepository().setURI(plugin.getScm()).setDirectory(tempDirectory.toFile()).call()) {
+                        final RevCommit head = new RevWalk(git.getRepository()).parseCommit(ObjectId.fromString("HEAD"));
+                        final ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(
+                            head.getAuthorIdent().getWhenAsInstant(),
+                            head.getAuthorIdent().getZoneId()
+                        );
+                        return ProbeResult.success(key(), zonedDateTime.toString());
+                    } finally {
+                        Files.delete(tempDirectory);
```
dheerajodha, line 51: Please correct me if I'm wrong, but I'm observing that this method does not delete the plugin directory (and all of its files and subdirectories)  and I'm hitting this exception: `java.nio.file.DirectoryNotEmptyException: /tmp/job-log-logger-plugin6397033103289588371` for example. Due to this, the status of this probe results for each plugin in the DB is a failure. Do you see this as well?

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
dheerajodha: I thought that we wouldn't need to store anything for this probe in DB if the previous probe is null.
But I can see why that might be confusing for us in the future when we don't see any results for this probe in DB, would be difficult for us to debug what went wrong. But with `ProbeResult.failure` we'd capture this message in DB.

Sorry, I should have asked you this question in the Probe Engine PR, but when would you use `ProbeResult.failure` VS `ProbeResult.error`? Because in this case, I thought it was an error for the probe if the previous probe is null.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
dheerajodha: I think I tried that and it did not delete the content of the directory and the directory itself as well.
So, I included a custom `deleteDirectory` method to recursively the contents of the given directory and also delete the given directory itself in the end.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
alecharp: More than just that, it is not able to get the object id `HEAD`, so I'm working on it.

dheerajodha: It works if we pass `git.getRepository().resolve(Constants.HEAD)` to the `parseCommit()`, but I'm not sure if this is the ideal way to do it.

alecharp: this is how I did it.

dheerajodha: Just saw your commits :+1:

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
alecharp: temp files are normally deleted when you restart your computer.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
alecharp: You are correct. Let's keep it this way.

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
dheerajodha: I initially tried to mock the Classes and got a little confused then I thought "Why do I even need to mock if I can use demo values as per the test case". So I took a shortcut without realizing that it isn't an ideal way to write unit tests and didn't consider the points you raised about the internet being down.

And then I forgot to delete this annotation.

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
dheerajodha: These points helped me understand why we even bother mocking the classes, to remove as many dependencies as possible, and just test the given class in question.

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
dheerajodha: Thanks!

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
```diff
@@ -0,0 +1,86 @@
...
+
+            if (plugin.getDetails().get(SCMLinkValidationProbe.KEY).status() == ResultStatus.SUCCESS) {
+                try {
+                    final Path tempDirectory = Files.createTempDirectory(plugin.getName());
+                    try (Git git = Git.cloneRepository().setURI(plugin.getScm()).setDirectory(tempDirectory.toFile()).call()) {
+                        final ObjectId head = git.getRepository().resolve(Constants.HEAD);
+                        final RevCommit commit = new RevWalk(git.getRepository()).parseCommit(head);
+                        final ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(
+                            commit.getAuthorIdent().getWhenAsInstant(),
+                            commit.getAuthorIdent().getZoneId()
```
dheerajodha, line 52: Ok so now you're converting the commit time as per the timezone the author is in, rather than using the default system timezone which varies (as you told me earlier).

1. Also, you replaced `.getWhen().toInstant()` with `.getWhenAsInstant()`. How did you know about this method `.getWhenAsInstant()`?
    * Did you already use it in past? and just knew about it already.
    * Or did you just type `commit.getAuthorIdent().getWhen` and read through all the possible methods available as shown by IntelliJ and noticed that there's already a method that returns `Instant`?

alecharp: You need to look for javadoc, explore the classes you are using and all. This way you can discover handy method like that.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
dheerajodha: This is not the only reason to get this exception. I should move this catch lock to the exact try block from where this exception could occur. I should also change the message to something which tells us that we have hit an exception related to Git API. Working on it.

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
```diff
@@ -0,0 +1,83 @@
...
+        final Plugin plugin = mock(Plugin.class);
+        final LastCommitDateProbe probe = new LastCommitDateProbe();
+
+        when(plugin.getDetails()).thenReturn(Map.of(SCMLinkValidationProbe.KEY, ProbeResult.success("scm", "The plugin SCM link is valid")));
+        when(plugin.getScm()).thenReturn("https://github.com/jenkinsci/parameterized-trigger-plugin.git");
+        final ProbeResult r = probe.apply(plugin);
+
+        assertThat(r.id()).isEqualTo("last-commit-date");
+        assertThat(r.status()).isEqualTo(ResultStatus.SUCCESS);
+    }
```
alecharp, line 60: Because this would validate the deletion of the file and all, I think it would be better to have an integration test.

dheerajodha: Sure, I'm on it.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
```diff
@@ -0,0 +1,85 @@
...
+                tempDirectory = Files.createTempDirectory(plugin.getName());
+                try (Git git = Git.cloneRepository().setURI(plugin.getScm()).setDirectory(tempDirectory.toFile()).call()) {
+                    final ObjectId head = git.getRepository().resolve(Constants.HEAD);
+                    final RevCommit commit = new RevWalk(git.getRepository()).parseCommit(head);
+                    final ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(
+                        commit.getAuthorIdent().getWhenAsInstant(),
+                        commit.getAuthorIdent().getZoneId()
+                    );
+                    return ProbeResult.success(key(), zonedDateTime.toString());
+                } catch (GitAPIException ex)  {
```
dheerajodha, line 55: We'll come into this catch block when there's a problem in the `call()` method above, while we try to clone the plugin repo.

I ran this on DB and it is quite helpful because now I see that there are some plugins whose status for `SCMLinkValidationProbe` is a SUCCESS, but for this probe, the status is FAILURE with a message `"Could not clone the plugin repository"`.

For example, there are some plugins (https://github.com/jenkinsci/aws-java-sdk-plugin/aws-java-sdk-logs) with valid SCM as per the GH_REGEXP but we get 404 (as per my understanding, this plugin should fail SCMLinkValidationProbe too) on its SCM link so this probe fails to clone it.

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
```diff
@@ -0,0 +1,91 @@
...
+        final LastCommitDateProbe probe = new LastCommitDateProbe();
+        assertNull(probe.tempDirectory);
+
+        when(plugin.getDetails()).thenReturn(Map.of(SCMLinkValidationProbe.KEY, ProbeResult.success("scm", "The plugin SCM link is valid")));
+        when(plugin.getScm()).thenReturn("https://github.com/jenkinsci/parameterized-trigger-plugin.git");
+        when(plugin.getName()).thenReturn("parameterized-trigger");
+        final ProbeResult r = probe.apply(plugin);
+
+        assertNotNull(probe.tempDirectory);
+        assertFalse(Files.exists(probe.tempDirectory));
```
dheerajodha, line 65: Just out of curiosity, do we really need an integration test to check if the plugin folder gets created and then gets deleted in the end or not?

Because here I'm asserting at the beginning of the tempDirectory for the current probe is null (Line 57), and it's non-null after we have run `probe.apply()` (Line 64) signifying that a tempDirectory has been created. And then I assert False to the existence of any folder at the `probe.tempDirectory` path (Line 65), signifying the folder has been deleted.

This test fails if I comment out the code which deletes the plugin directory.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
alecharp: technically, the SCM is valid because the plugin is a submodule of the repository. But it is true that it is not a valid link to clone the repository.

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeTest.java
alecharp: `tempDirectory` should definitively **not** be public. 
This is the case in 99% of the situations: class field should **not** be public.

src/main/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbe.java
alecharp: that is returning a `404`? `SCMLinkValidationProbe`? But still have a `success`?

dheerajodha: image
I see the status code as `404` under the "Status" tab on the bottom right. I thought that's the same status code we're checking via `response.statusCode()`

alecharp: That is normal. The `SCMLinkValidationProbe` is not taking into account submodules to validate the repository. This means our code coverage is not enough. Working on it.

dheerajodha: > technically, the SCM is valid because the plugin is a submodule of the repository.

So now I understand what you mean by "plugin is a submodule of the repository". If I visit this link: https://github.com/jenkinsci/aws-java-sdk-plugin/ I can see there are "submodules" like `-ec2`, `-cs`, `-efs` etc. And these submodules are a plugin on their own (I might be wrong here) that's why they are distributed separately in the update-center.

> But it is true that it is not a valid link to clone the repository.

So, I guess there's no separate SCM Link to clone each of these submodules because that doesn't make sense as all of them reside in a single repo.

alecharp: > So now I understand what you mean by "plugin is a submodule of the repository".

it is a Maven module, not a Git submodule (which also is a existing concept). Sorry if I made things more complex to understand.

> And these submodules are a plugin on their own (I might be wrong here) that's why they are distributed separately in the update-center.

Correct.

> So, I guess there's no separate SCM Link to clone each of these submodules because that doesn't make sense as all of them reside in a single repo.

There is, but I'm not sure JGit can clone only a part of the repository. I'm investigating.

src/test/java/io/jenkins/pluginhealth/scoring/probes/LastCommitDateProbeIT.java
```diff
@@ -0,0 +1,28 @@
...
+        final Plugin plugin = new Plugin("parameterized-trigger", "https://github.com/jenkinsci/parameterized-trigger-plugin.git", ZonedDateTime.now())
+            .addDetails(ProbeResult.success("scm", "The plugin SCM link is valid"));
+        final LastCommitDateProbe probe = new LastCommitDateProbe();
+
+        final ProbeResult r = probe.apply(plugin);
+
+        assertThat(r.id()).isEqualTo("last-commit-date");
+        assertThat(r.status()).isEqualTo(ResultStatus.SUCCESS);
+    }
+}
```
dheerajodha, line 28: @alecharp I'm very sorry for delaying this PR for the Integration test you asked for 2 days ago. I see that you had to mark that comment as resolved because I didn't commit anything to it since then.

An integration would run everything from start to end, with no mocks at all. So, for that I have added `extends AbstractDBContainerTest` to this class so that we have a running container while we run this test. The aim of this integration test was to validate whether the cloned repo is getting deleted in the end or not.

So, I just simply copy pasted the unit test here in this class with dummy values. But my issue is how do I access the path of this plugin's repo inside this class? I'm asking this because, when I have access to the path of the repo, I would be able to use the method like `Files.exists()` to assert the absence of any data at the cloned-repo's path, and that would solve what we want to validate.

alecharp: sorry I was silent on this subject. 
let's not bring an integration test for the moment. The unit ones are already checking this. We will see down the road of the project if we have problems with this portion of code.

dheerajodha: Sounds good, thank you

dheerajodha:
I'm now looking into why the test is failing.

PR #36, opened 2022-09-02
Application deployment

alecharp:
This enable CD for the application to be deployed on Jenkins project infra.
Related to https://github.com/jenkins-infra/helpdesk/issues/3114

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,9 @@
 #!/usr/bin/env groovy
 
 pipeline {
-  agent none
+  agent {
+    label 'docker && linux'
+  }
   options {
     buildDiscarder(logRotator(numToKeepStr: '10'))
     timestamps()
@@ -9,9 +11,6 @@ pipeline {
 
   stages {
     stage('Build') {
-      agent {
-        label 'linux'
-      }
       environment {
         JAVA_HOME = '/opt/jdk-17/'
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,16 +1,17 @@
 #!/usr/bin/env groovy
 
 pipeline {
-  agent {
-    label 'docker && linux'
-  }
+  agent none
   options {
-    buildDiscarder(logRotator(numToKeepStr: '5'))
+    buildDiscarder(logRotator(numToKeepStr: '10'))
     timestamps()
   }
 
   stages {
     stage('Build') {
+      agent {
+        label 'linux'
+      }
       environment {
         JAVA_HOME = '/opt/jdk-17/'
       }
@@ -33,6 +34,24 @@ pipeline {
           recordIssues enabledForFailure: true, tool: checkStyle()
           recordIssues enabledForFailure: true, tool: spotBugs()
         }
+        success {
+            stash name: 'binary', includes: 'target/plugin-health-scoring.jar,src/main/docker/Dockerfile'
+        }
+      }
+    }
+
+    stage('Docker image') {
+      /* when {
+          branch 'main' // TODO for now, activated for the PR
+      } */
+      steps {
+        unstash 'binary'
+        buildDockerAndPublishImage(
+          imageShortName: 'plugin-health-scoring',
+          userConfig: [
+            dockerfile: 'src/main/docker/Dockerfile',
+          ]
+        )
       }
     }
   }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,8 +45,7 @@ pipeline {
           branch 'main' // TODO for now, activated for the PR
       } */
       steps {
-        unstash 'binary'
-        buildDockerAndPublishImage('plugin-health-scoring', [dockerfile: 'src/main/docker/Dockerfile', skipCheckout: true])
+        buildDockerAndPublishImage('plugin-health-scoring', [dockerfile: 'src/main/docker/Dockerfile', unstash: 'binary'])
       }
     }
   }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,12 +45,7 @@ pipeline {
       } */
       steps {
         unstash 'binary'
-        buildDockerAndPublishImage(
-          imageShortName: 'plugin-health-scoring',
-          userConfig: [
-            dockerfile: 'src/main/docker/Dockerfile',
-          ]
-        )
+        buildDockerAndPublishImage('plugin-health-scoring', [dockerfile: 'src/main/docker/Dockerfile'])
       }
     }
   }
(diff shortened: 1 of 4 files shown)


lemeurherve:
The `plugin-health-scoring` image must be created in the docker hub before being able to push on it, looking into it.

alecharp:
Don't worry. I'm preparing the ground here, so I expect the build to fail for now.

Jenkinsfile
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env groovy
+@Library('pipeline-library@pull/465/head') _
```
alecharp, line 2: Will be removed once https://github.com/jenkins-infra/pipeline-library/pull/465 is merged.

Jenkinsfile
```diff
@@ -33,6 +34,18 @@ pipeline {
...
+        success {
+            stash name: 'binary', includes: 'target/plugin-health-scoring.jar,src/main/docker/Dockerfile'
+        }
+      }
+    }
+
+    stage('Docker image') {
+      /* when {
+          branch 'main' // TODO for now, activated for the PR
+      } */
```
alecharp, line 46: will be uncommented before being merged.

Jenkinsfile
```diff
@@ -33,6 +34,18 @@ pipeline {
           recordIssues enabledForFailure: true, tool: checkStyle()
           recordIssues enabledForFailure: true, tool: spotBugs()
         }
+        success {
+            stash name: 'binary', includes: 'target/plugin-health-scoring.jar,src/main/docker/Dockerfile'
```
lemeurherve, line 38:
```suggestion
            stash name: 'binary', includes: 'target/plugin-health-scoring.jar'
```
After discussion with @dduportal [I've modified my pipeline library PR](https://github.com/jenkins-infra/pipeline-library/pull/465/commits/652268cdcae3e510fe9f39f6ce1a9ceec11e7e2e) in order to be more generic by doing the `checkout` even if `unstash` isn't empty. (To avoid mistake by other users)
Thus, stashing only the resulting .jar is enough here.

Jenkinsfile
dduportal: I would remove it instead:

- There is already a condition for publishing the image in the library: https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildDockerAndPublishImage.groovy#L205
- It means that you are going to lint + build (+eventually test) the Docker image for each change, which make sense as it's the final artefact (so the application would always be deployable for each commit with successfull check)

WDYT?

compose.yml
```diff
@@ -13,6 +13,8 @@ services:
         condition: service_healthy
     volumes:
       - db_data:/var/lib/postgresql/data
```
dduportal, line 15: nitpick(unrelated to this PR): this volume mount looks weird:

- There is no ["top level `volumes` directive"](https://docs.docker.com/compose/compose-file/compose-file-v3/#volume-configuration-reference) referencing the named volume `db_data`
- It's not clear why the application would have direct access to postgres "raw" data on the file system

Did I miss something?

compose.yml
```diff
@@ -13,6 +13,8 @@ services:
         condition: service_healthy
```
dduportal, line 13: nitpick(unrelated to this PR):  using the `service_healthy` is a really good idea! But healthchecks are required on the "target" dependency, otherwise Docker considers that `service_healthy === service_started`.

In this particular case, the `db` service does not have an healthcheck defined: https://github.com/docker-library/postgres/blob/56a1986772dd0f9488d54dccb82427c0db0b0599/14/alpine/Dockerfile
`
But you can add one to the `db` service through docker-compose to make this really strong and keep the initial intent (e.g. only starting the Java app once you're assurred that the DB service is up) with this directive: https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck .

The rationale behind "NOT" defining a default healthcheck for official image is because the "healthcheck" is a Docker-only feature: it will be ignored on Kubernetes, containerd, podman, etc. 
Kubernetes, for instance, has the concept of "probes" instead, which are fundamentally different: Kubernetes porbes are executed by a process *outside* the container (`kubelet`), while the Docker healthcheck are executed *inside* the container (hence the need to make sure that `curl`, or `wget` or any tools is installed in your image).

dduportal: 🤔 I'm thinking a bit in advance but this comment might not be so much nitpick:

- The target is Kubernetes for the application. But we'll most probably use a postgres managed database outside Kubernetes (e.g. not run in a container, neither in the same orchestrator).

=> so maybe you want to remove the `depends_on` directive here at all (*when* you'll be ready, of course :)) to simulate what could happen in production: the application container being up while the postgres database is not started or reachable for a few seconds.
We expect the application to crash (most probably), so a ['restart' policy](https://docs.docker.com/compose/compose-file/compose-file-v3/#restart) could be required.

Jenkinsfile
```diff
@@ -33,6 +34,18 @@ pipeline {
...
+        success {
+            stash name: 'binary', includes: 'target/plugin-health-scoring.jar'
+        }
+      }
+    }
+
+    stage('Docker image') {
+      /* when {
+          branch 'main' // TODO for now, activated for the PR
+      } */
```
alecharp, line 46:
```suggestion
```

compose.yml
alecharp: https://github.com/jenkins-infra/plugin-health-scoring/pull/36/files#diff-3493e6b5ddf34891e572f911db893efd9e46af828e011ea778a7c1eb64763588R24-R28

compose.yml
alecharp: Yes, this is something I forgot to declare here, and that is why the db content is erased at each restart of the db service.

compose.yml
dduportal: as pointed by @alecharp , there is already an health check on the db: I forgot to unfold 🤦 

Sorry for the noise

pom.xml
```diff
@@ -106,6 +106,11 @@
 			<artifactId>thymeleaf-layout-dialect</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-actuator</artifactId>
```
dduportal, line 111: Note for infra-team: https://www.baeldung.com/spring-boot-actuators , it provides probes for Docker healthcheck/Kubernetes probes.
Thanks @alecharp !

Jenkinsfile
alecharp: This is a very good point. I applied your suggestion.

dduportal (approved):
👋

PR #38, opened 2022-09-08
Adds structure for update-center-related probes

### LINKED ISSUE #34: Details coming from Update Center
In the `update-center.json` file, there are a lot of details that we might be interested in. 
However, this file is only parsed when we update the database for the list of plugins. Because of this, the `UpdateCenterService` is less and less about the update-center. 

We need to create a structure of probes related to the update-center, which would download the file once, parse it, and pass its representation to the different probes to gather some details from.

This will open the door to probes like: 
 - minimum version of Jenkins required for the plugin
 - is the plugin up for adoption
 - etc.

dheerajodha:
Closes #34


src/main/java/io/jenkins/pluginhealth/scoring/probes/CompoundUpdateCenterProbes.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.core.annotation.Order;
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(value = CompoundUpdateCenterProbes.ORDER)
+public class CompoundUpdateCenterProbes extends Probe implements UpdateCenterProbe {
```
dheerajodha, line 17: This is `Container (aka composite)` based on the composite design pattern. Ideally, it should:
 >  download the update-center.json file and parse it. Then, it would call other classes which would take the Plugin and the UpdateCenter classes in a method as argument.

But my implementation of `readUpdateCenter` below is wrong and doesn't make any difference because each probe reads the update center from the scratch by itself. I couldn't make sense of how I could pass the `UpdateCenter` classes in this method as an argument.

src/main/java/io/jenkins/pluginhealth/scoring/probes/CompoundUpdateCenterProbes.java
```diff
@@ -0,0 +1,53 @@
...
+    protected ProbeResult doApply(Plugin plugin) {
+        try {
+            readUpdateCenter(plugin);
+
+            return ProbeResult.success(key(), "Read the update center for all the probes");
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return ProbeResult.failure(key(), "Error in reading the update-center");
+    }
```
dheerajodha, line 47: Also since I'm using `extends Probe` in this class, I was wondering what should I return from this `doApply` method.
This is adding a new probe in the `details` of the plugin, but this isn't a probe.

I made this a probe because of this line:
> I'll have one probe which downloads the update-center.json file and parse it.

I thought about passing the representation of the `plugins` object from the `update-center.json` file, but couldn't understand how using this object would the update-center-related probes could fetch its inner objects.

alecharp (changes requested):
With this implementation, the `update-center.json` file is downloaded by each probe.

Because the new probes are extending `Probe`, aren't they also executed directly by the ProbeEngine?

src/main/java/io/jenkins/pluginhealth/scoring/probes/DeprecatedPluginProbe.java
```diff
@@ -0,0 +1,55 @@
...
+
+    record UpdateCenterDeprecations(String url) {
+    }
+
+    record UpdateCenter(Map<String, UpdateCenterDeprecations> deprecations) {
+    }
+
+    @Override
+    public void readUpdateCenter(Plugin plugin) throws IOException {
+        updateCenter = objectMapper.readValue(new URL(updateCenterURL), UpdateCenter.class);
```
dheerajodha, line 40: Ideally, I didn't want to read the update center here, because the `CompoundUpdateCenterProbes` class would have already read it once and passed it's representation to this probe via this method.
And then I just wanted to use it to extract details related to deprecations. But I don't know how I can do this.

src/main/java/io/jenkins/pluginhealth/scoring/probes/DeprecatedPluginProbe.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin) {
+        if (updateCenter.deprecations().containsKey(plugin.getName())) {
+            return ProbeResult.failure(key(), updateCenter.deprecations().get(plugin.getName()).url());
+        }
+        return ProbeResult.success(key(), "This plugin is NOT deprecated");
+    }
```
dheerajodha, line 49: Is this correct?
We are not reading the `update-center.json` file inside the `doApply` method, instead, we're using the `updateCenter` object which contains the representation which was previously read.

dheerajodha (commented):
I have done changes that are inefficient, in an attempt to use the composite design pattern in this, but I couldn't clearly understand how it could be used.

src/main/java/io/jenkins/pluginhealth/scoring/probes/CompoundUpdateCenterProbes.java
alecharp: The `doApply(Plugin)` would download and parse the file and then call another method on each subprobes with both the `UpdateCenter` and the `Plugin` objects.

src/main/java/io/jenkins/pluginhealth/scoring/probes/UpdateCenterProbe.java
```diff
@@ -0,0 +1,9 @@
+package io.jenkins.pluginhealth.scoring.probes;
+
+import java.io.IOException;
+
+import io.jenkins.pluginhealth.scoring.model.Plugin;
+
+public interface UpdateCenterProbe {
```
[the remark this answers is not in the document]
alecharp: its purpose shouldn't be to read the `update-center.json` file but to analyze its Java representation with a `Plugin` class.

src/main/java/io/jenkins/pluginhealth/scoring/probes/DeprecatedPluginProbe.java
alecharp: the idea is to only download and parse the `update-center.json` once for all the probes that requires it.

dheerajodha: How would I parse the `update-center.json` just once at the beginning for all the probes that require it?

The `readValue` method's second argument is taking the Class which it will use as a blueprint to extract data from the JSON object.

And since the data-to-be-extracted varies from probe to probe (probe related to deprecations doesn't need "plugins" object, probe related to adoption needs the object "plugins"->"labels"), I'm confused about which class should I pass as the 2nd arg to the `readValue` method? Currently, we're using `records` as an ad-hoc replacement for POJO, but I don't think that would be the way to go in this scenario.

alecharp: I'm working on this PR to match this and even improve #32.

src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngine.java
```diff
@@ -47,19 +50,23 @@ public class ProbeEngine {
...
     /**
      * Starts to apply all the {@link Probe} implementations on all the plugins registered in the database.
      */
-    public void run() {
+    public void run() throws IOException {
         LOGGER.info("Start running probes on all plugins");
+        final UpdateCenter updateCenter = updateCenterService.fetchUpdateCenter();
         pluginService.streamAll().parallel()
             .peek(plugin -> {
+                final ProbeContext probeContext = ProbeContext.withUpdateCenter(updateCenter);
```
dheerajodha, line 69: Optional question, answer only if you have time (please feel free to ignore):

1. Why is the `withUpdateCenter()` static?
What if it wasn't static and I called it from here like this: `final ProbeContext probeContext = new ProbeContext(updateCenter);`? Did you make it static **only** to avoid the `May expose internal representation ...` error?

2. Why does each plugin gets a new separate instance of the `probeContext` object?
As per my understanding, all the update-center-related probes are just using this context to access the update center's representation. Wouldn't this still work the same way if we just instantiate this object once?

alecharp: > 1.
Because I found it easier at the time.

> 2. 
Because at some point, the context will hold more that just the update-center representation but intermediate state of the probes executions to carry details from probe to probe without putting that into a `ProbeResult`

alecharp (approved):
Should be good enough to be merged now

PR #41, opened 2022-09-20
feat: build on infra.ci.jenkins.io too

lemeurherve:
Agent label changed to one existing on both ci.jenkins.io and infra.ci.jenkins.io

Complement of https://github.com/jenkins-infra/kubernetes-management/pull/2889

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@
 
 pipeline {
   agent {
-    label 'docker && linux'
+    label 'linux-amd64-docker'
   }
   options {
     buildDiscarder(logRotator(numToKeepStr: '10'))
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,7 +38,7 @@ pipeline {
         }
         success {
             stash name: 'binary', includes: 'target/plugin-health-scoring.jar'
-            stash name: 'extractedBinaries', includes: 'target/application/**', 'target/dependencies/**', 'target/snapshot-dependencies/**', 'target/spring-boot-loader/**'
+            stash name: 'extractedBinaries', includes: 'target/application/**, target/dependencies/**, target/snapshot-dependencies/**, target/spring-boot-loader/**'
         }
       }
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,6 +18,7 @@ pipeline {
       }
       steps {
         sh './mvnw -V verify checkstyle:checkstyle spotbugs:spotbugs -Dmaven.test.failure.ignore -Dcheckstyle.failOnViolation=false -Dspotbugs.failOnError=false'
+        sh 'cd target && java -Djarmode=layertools -jar plugin-health-scoring.jar extract'
       }
 
       post {
@@ -37,13 +38,14 @@ pipeline {
         }
         success {
             stash name: 'binary', includes: 'target/plugin-health-scoring.jar'
+            stash name: 'extractedBinaries', includes: 'target/application/**', 'target/dependencies/**', 'target/snapshot-dependencies/**', 'target/spring-boot-loader/**'
         }
       }
     }
 
     stage('Docker image') {
       steps {
-        buildDockerAndPublishImage('plugin-health-scoring', [dockerfile: 'src/main/docker/Dockerfile', unstash: 'binary'])
+        buildDockerAndPublishImage('plugin-health-scoring', [dockerfile: 'src/main/docker/Dockerfile', unstash: 'extractedBinaries'])
       }
     }
   }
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -3,7 +3,7 @@
 
 pipeline {
   agent {
-    label 'docker && linux'
+    label 'linux-amd64-docker'
```
dduportal, line 6:
```suggestion
    // 'docker' is the (legacy) label used on ci.jenkins.io for "Docker Linux AMD64" while 'linux-amd64-docker' is the label used on infra.ci.jenkins.io
    label 'docker || linux-amd64-docker'
```

dduportal (changes requested):

Jenkinsfile
hervelemeur: Isn't there a risk of getting a Windows Docker agent with `docker` on ci.jenkins.io?

dduportal: Good question! But the answer is no:
- Azure Windows VM labels on ci.jenkins.io: https://github.com/jenkins-infra/jenkins-infra/blob/production/hieradata/clients/azure.ci.jenkins.io.yaml#L346
- EC2 Windows VM labels on ci.jenkins.io: https://github.com/jenkins-infra/jenkins-infra/blob/production/hieradata/clients/azure.ci.jenkins.io.yaml#L428


=> we still have to update the label policy to a "safe" on on ci.jenkins.io, but right now, `docker` is safe (thanks to your work as far as I remember)

alecharp (approved):

dheerajodha:
[Please correct me if I'm wrong]:

The intention behind this PR is to allow the CI process of this project to run on both `ci.jenkins.io` and `infra.ci.jenkins.io`.

Currently, the CI is successfully running on `ci.jenkins.io`, and to make it run on `infra.ci.jenkins.io` we need to add the acceptable agent label for both the services.

And we're extending our CI to `infra.jenkins.io` because we want to publish the project's container image to the `jenkinsciinfra/plugin-health-scoring` repository which needs the CI job to execute on `infra.ci.jenkins.io`.

lemeurherve:
> The intention behind this PR is to allow the CI process of this project to run on both `ci.jenkins.io` and `infra.ci.jenkins.io`.

Exactly.

> Currently, the CI is successfully running on `ci.jenkins.io`, and to make it run on `infra.ci.jenkins.io` we need to add the acceptable agent label for both the services.

The previous label `docker && linux` applied for ci.jenkins.io didn't had any match in the available agents on infra.ci.jenkins.io, I replaced it with the one used in `buildDockerAndPublishImage` (as this shared pipeline function spawn its own agent), available on both ci.jenkins.io and infra.ci.jenkins.io

> And we're extending our CI to `infra.jenkins.io` because we want to publish the project's container image to the `jenkinsciinfra/plugin-health-scoring` repository which needs the CI job to execute on `infra.ci.jenkins.io`.

You've understood it correctly 🙂

lemeurherve:
Note: I've simplified the build process as there was an error retrieving the new folders from the `builder` Docker layer on infra.ci.jenkins.io

Feel free to reimplement a multi-layer Dockerfile later.

src/main/docker/Dockerfile
```diff
@@ -1,20 +1,13 @@
...
-COPY --from=builder /app/dependencies/ ./
-COPY --from=builder /app/spring-boot-loader/ ./
-COPY --from=builder /app/snapshot-dependencies/ ./
-COPY --from=builder /app/application/ ./
+COPY target/application/ ./
+COPY target/dependencies/ ./
+COPY target/snapshot-dependencies/ ./
+COPY target/spring-boot-loader/ ./
 
 ENTRYPOINT ["java","org.springframework.boot.loader.JarLauncher"]
```
dduportal, line 13:
```suggestion
COPY target/plugin-health-scoring.jar /app/plugin-health-scoring.jar

ENTRYPOINT ["java", "-jar", "/app/plugin-health-scoring.jar"]
```

the layerisation of jar is nice, but it is an optimization that we do not really need. Since it's failing, let's keep a "classic" Dockerfile model and we'll improve later (tip: multi-stage should use the same JRE, and we use temurin instead of azul + openjdk is deprecated in favor of temurin).

Jenkinsfile
```diff
@@ -17,6 +18,7 @@ pipeline {
       }
       steps {
         sh './mvnw -V verify checkstyle:checkstyle spotbugs:spotbugs -Dmaven.test.failure.ignore -Dcheckstyle.failOnViolation=false -Dspotbugs.failOnError=false'
+        sh 'cd target && java -Djarmode=layertools -jar plugin-health-scoring.jar extract && echo "--------- LS -------------" && ls .'
```
dduportal, line 21:
```suggestion
```

se my other comment below about starting simple

src/main/docker/Dockerfile
lemeurherve: I don't understand why the application and dependencies folders are copied without issue while the snapshot-dependencies one is failing. Is it because of the folder name?

dduportal (approved):

PR #64, opened 2022-10-06
Jenkinsfile existence probe

### LINKED ISSUE #54: New probe: Does the plugin has a `Jenkinsfile`?
Having a `Jenkinsfile` means, for a plugin, that the CI is enabled and any pull requests is built. This is important for maintainability and to welcome contributions on the plugins.

alecharp:
Resolves #54 by checking the plugin repository, cloned by the `LastCommitDateProbe` in the context scm location, to see if it has a `Jenkinsfile` file at the root level.

### DIFF
--- a/src/main/java/io/jenkins/pluginhealth/scoring/probes/JenkinsfileProbe.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/probes/JenkinsfileProbe.java
@@ -1,0 +1,45 @@
+package io.jenkins.pluginhealth.scoring.probes;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.stream.Stream;
+
+import io.jenkins.pluginhealth.scoring.model.Plugin;
+import io.jenkins.pluginhealth.scoring.model.ProbeResult;
+
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(value = JenkinsfileProbe.ORDER)
+public class JenkinsfileProbe extends Probe {
+    public static final int ORDER = LastCommitDateProbe.ORDER + 100;
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        final Path repository = context.getScmRepository();
+        try (Stream<Path> paths = Files
+            .find(repository, 1, (file, basicFileAttributes) -> Files.isReadable(file) && "Jenkinsfile".equals(file.getFileName().toString()))
+        ) {
+            return paths.findFirst()
+                .map(file -> ProbeResult.success(key(), "Jenkinsfile found"))
+                .orElseGet(() -> ProbeResult.failure(key(), "No Jenkinsfile found"));
+        } catch (IOException e) {
+            return ProbeResult.error(key(), e.getMessage());
+        }
+    }
+
+    @Override
+    public String key() {
+        return "jenkinsfile";
+    }
+
+    @Override
+    public String getDescription() {
+        return """
+    Validates the existence of a `Jenkinsfile` file on the repository.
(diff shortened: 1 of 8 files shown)


src/main/java/io/jenkins/pluginhealth/scoring/probes/JenkinsfileProbe.java
```diff
@@ -0,0 +1,45 @@
...
+@Component
+@Order(value = JenkinsfileProbe.ORDER)
+public class JenkinsfileProbe extends Probe {
+    public static final int ORDER = LastCommitDateProbe.ORDER + 100;
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        final Path repository = context.getScmRepository();
+        try (Stream<Path> paths = Files
+            .find(repository, 1, (file, basicFileAttributes) -> Files.isReadable(file) && "Jenkinsfile".equals(file.getFileName().toString()))
```
hervelemeur, line 23: Nit: I would search for `Jenkinsfile*`, there are a few exceptions.

alecharp: I know that there are google plugins (ex: https://github.com/jenkinsci/google-kubernetes-engine-plugin) which have a `Jenkinsfile.google` used for their infra. If that is the only `Jenkinsfile` present on the repository, do we still want to consider those plugins as ok?

alecharp: Do you have a lot of example? Because when I run this code, I found that 714 plugins don't have `Jenkinsfile` and when I change it to find files starting with `Jenkinsfile` in their name, I found that 712 plugins don't have that. So, no big difference really..

lemeurherve: As we have some `Jenkinsfile_k8s` and other exotic filenames in the jenkins-infra org, hence my suggestion, but it seems we are the exception.
We can resolve this conversation, ignore my suggestion please.

src/main/java/io/jenkins/pluginhealth/scoring/probes/JenkinsfileProbe.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public String key() {
+        return "jenkinsfile";
+    }
+
+    @Override
+    public String getDescription() {
+        return """
+    Validates the existence of a `Jenkinsfile` file on the repository.
```
lemeurherve, line 41:
```suggestion
    Validates the existence of a `Jenkinsfile` file in the repository.
```

src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeEngine.java
```diff
@@ -66,31 +66,36 @@ public void run() throws IOException {
...
+                                if (LOGGER.isTraceEnabled()) {
+                                    LOGGER.trace("Running {} on {}", probe.key(), plugin.getName());
+                                }
+                                final ProbeResult result = probe.apply(plugin, probeContext);
+                                if (result.status() != ResultStatus.ERROR) {
+                                    plugin.addDetails(result);
+                                }
+                            } else {
+                                if (LOGGER.isDebugEnabled()) {
+                                    LOGGER.debug("{} requires a release of {} to process it again.", probe.key(), plugin.getName());
```
lemeurherve, line 88:
```suggestion
                                    LOGGER.debug("{} requires a release of {} to be processed.", probe.key(), plugin.getName());
```
Not sure about this one.

lemeurherve (approved):
(nit only)

lemeurherve:
Erf, too late.

alecharp:
Sorry, I didn't wait.. let's create a PR with those suggestions.

PR #66, opened 2022-10-07
Checks update-center for known vulnerabilities on plugins

### LINKED ISSUE #55: New probe: Does the plugin has known security vulnerability?
There are disclosed security vulnerability and the security team is advertising them on the `update-center.json` file. 
Having a public security vulnerability is a big deal for a plugin. 

This new probe should detect plugins which such warnings, determine if the current release available in the Update Center is impacted by it and is so log that.

alecharp:
Resolves #55 by looking at the `warnings` section of the `update-center`. In this section we can find publicly known security vulnerabilities, so we can determine if they apply to the current plugin and to its current version.

In order to get the required details, I had to retrieve the plugin latest version from the update-center and store it in the DB. 

The trick is that, a plugin *requires* a release to not be impacted by a security vulnerability, but can be impacted by a new one without a new release. So this probe needs to be ran every time.

### DIFF
--- a/src/main/java/io/jenkins/pluginhealth/scoring/probes/KnownSecurityVulnerabilityProbe.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/probes/KnownSecurityVulnerabilityProbe.java
@@ -44,6 +44,6 @@ public class KnownSecurityVulnerabilityProbe extends Probe {
 
     @Override
     public String getDescription() {
-        return "Detects if a given plugin as a public security vulnerability advertised in a security advisory.";
+        return "Detects if a given plugin has a public security vulnerability advertised in a security advisory.";
     }
 }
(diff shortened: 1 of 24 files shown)


src/main/java/io/jenkins/pluginhealth/scoring/probes/KnownSecurityVulnerabilityProbe.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public String key() {
+        return "security";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Detects if a given plugin as a public security vulnerability advertised in a security advisory.";
```
lemeurherve, line 47:
```suggestion
        return "Detects if a given plugin has a public security vulnerability advertised in a security advisory.";
```
Could a plugin has more than one vuln? In that case, could it increase even more their deprecation score?

lemeurherve (approved):

src/main/java/io/jenkins/pluginhealth/scoring/probes/KnownSecurityVulnerabilityProbe.java
alecharp: This is a good point. I can have more than one I guess (cc @daniel-beck @Wadeck) so yes, we could log all the issues it has.

daniel-beck: E.g. https://plugins.jenkins.io/websphere-deployer/

alecharp: normally https://github.com/jenkins-infra/plugin-health-scoring/pull/66/commits/60c23e3b7fbe40ba851e028c7e86ddc656626ad0 is doing that.

PR #71, opened 2022-10-11
Validates documentation migration

### LINKED ISSUE #58: New probe: is the plugin documentation located on GitHub?
The Jenkins wiki has be retired some times ago. Since then, it is recommended for the plugins to use their `README` on their Git repository to hold their documentation. This is also used on `plugins.jenkins.io` to help users to use correctly the plugin.

This migration was not automatic and it is possible that some plugins are still missing it.

alecharp:
Resolves #58.

Based on https://www.jenkins.io/doc/developer/publishing/documentation guidelines. For the migration to be complete, it requires a `README.{md,adoc}` file and for the build tool to have the `url` configured.

Requires to parse build configuration file.

### DIFF
--- a/src/main/java/io/jenkins/pluginhealth/scoring/probes/DocumentationLocationProbe.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/probes/DocumentationLocationProbe.java
@@ -1,0 +1,77 @@
+package io.jenkins.pluginhealth.scoring.probes;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import io.jenkins.pluginhealth.scoring.model.Plugin;
+import io.jenkins.pluginhealth.scoring.model.ProbeResult;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.core.annotation.Order;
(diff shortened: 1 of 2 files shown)


alecharp:
At this moment, only Maven based projects can be handled. Working on Gradle ones

alecharp:
It seems that retrieving the extension configuration from the gradle project is not something that the gradle tooling library is providing (https://discuss.gradle.org/t/how-to-get-the-project-extensions-in-gradle-tooling/6042/4). 

The `gradle-jpi-plugin` is written in Groovy, Java and Kotlin. The extensions is registered here 
https://github.com/jenkinsci/gradle-jpi-plugin/blob/c813f75e64a409b65795f35eb79ede6982406e1b/src/main/groovy/org/jenkinsci/gradle/plugins/jpi/JpiPlugin.groovy#L103
and the `JpiExtension` would give us the required url (https://github.com/jenkinsci/gradle-jpi-plugin/blob/c813f75e64a409b65795f35eb79ede6982406e1b/src/main/groovy/org/jenkinsci/gradle/plugins/jpi/JpiExtension.groovy#L144-L146).

Based on https://github.com/search?l=Gradle&q=org%3Ajenkinsci+filename%3Abuild.gradle&type=Code, the ecosystem has 83 plugins built with Gradle.

alecharp:
For the record, the current lead to solve the problem with Gradle based plugin is to statically log them in a file on the project and use that file.

PR #82, opened 2022-10-13
Update Web UI UI/UX and tooling

alecharp:
Let's try to build the frontend of the application using dedicated tooling.
This comes with some modifications of the Web UI

<details>
<summary>Before</summary>

Screenshot 2022-10-14 at 16-03-22 Plugin Health Scoring Probes
</details>

<details>
<summary>After</summary>

Screenshot 2022-10-17 at 16-06-13 Plugin Health Scoring Probes

</details>


CONTRIBUTING.adoc
```diff
@@ -0,0 +1,93 @@
+= Contributing to Plugin Health Scoring
+
+This page provides information about contributing to Plugin Health Scoring code base.
```
kmartens27, line 3:
````
This page provides information about contributing to the Plugin Health Scoring code base.
````
small update for readability

File: CONTRIBUTING.adoc [hunk 2]
```diff
@@ -0,0 +1,93 @@
...
+
+== Getting started
+
+The project is using Maven with Java 17 and Yarn to be built.
+The entire build process can be handled by Maven.
+Yarn build will bundle the front-end files only.
```
kmartens27, line 9:
````
Yarn build will bundle the front-end files only.
````
When building, Yarn will bundle the front-end files only.

small suggestion (again, not necessary if it does not make the same sense)

File: CONTRIBUTING.adoc [hunk 3]
```diff
@@ -0,0 +1,93 @@
...
+
+. Clone the repository
```
kmartens27, line 11:
```
. Clone the repository
```
. Clone the repository.

adding a full stop since some of these items have a "."

CONTRIBUTING.adoc
```diff
@@ -0,0 +1,93 @@
...
+. Install tools
+.. Java 17.
+.. Maven 3.8.6.
+Optionally, you can use the wrapper in the repository.
+In that case, replace the `mvn` command by `./mvnw` in the command listed bellow.
+.. Node 18.10.0 and Yarn 3.2.4
+. Open the project in your favorite IDE
```
kmartens27, line 18:
````
. Open the project in your favorite IDE
````
. Open the project in your favorite IDE.
adding full stop here as well

File: CONTRIBUTING.adoc [hunk 5]
```diff
@@ -0,0 +1,93 @@
...
+The entire build process can be handled by Maven.
+Yarn build will bundle the front-end files only.
+
+. Clone the repository
+. Install tools
+.. Java 17.
+.. Maven 3.8.6.
+Optionally, you can use the wrapper in the repository.
+In that case, replace the `mvn` command by `./mvnw` in the command listed bellow.
+.. Node 18.10.0 and Yarn 3.2.4
```
kmartens27, line 17: Is line 17 the other tools that need to be installed? 

Would it change the message if lines 15 & 16 were placed below 17 instead? Or is Maven considered the default wrapper in this case?

CONTRIBUTING.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+== Proposing changes
+
+All proposed changes are submitted and reviewed through a GitHub pull request.
+To submit a pull request:
+
+. Create a branch prefixed with `feature/` for a new enhancement, `fix/` for a bugfix.
+. Commit your changes to this new branch.
+. Push the branch to your fork of the repository.
+. In GitHub Web UI, click the _New pull request_ button.
```
kmartens27, line 89:
```suggestion
. In GitHub Web UI, click the _New pull request_ button.
```
. In the GitHub Web UI, select the *New pull request* option.

Small suggestion, as we are trying to use select as opposed to click. Again, if this does not make sense, it does not need to be changed. Also, if the button is labeled as New pull request, it can be bolded so that users know what to look for.

kmartens27 (approved):
everything looks good @alecharp! I had a few suggestions on adding the `link:` macro and a couple of small wording suggestions, but approving as they are not mandatory

CONTRIBUTING.adoc
alecharp: Technically, you cannot do anything without Maven but Maven is not technically building everything either.. complex situation. 
The `wrapper` is only available to act as Maven and so contributors don't necessarily have to install it. The wrapper install it for them, in a sense. 

From a document syntax point of view, this list is rendered as

Screenshot 2022-10-14 at 15 39 40

So lines `15-16` are really in the list item started on line `14`.

kmartens27: Okay got it, that makes more sense. In that case everything here looks good to me!

alecharp:
`unpkg` is illegal in Europe it seems so I need to change how `ionicons` is loaded but found https://github.com/ionic-team/ionicons/issues/877 which is still not documented.

lemeurherve:
> `unpkg` is illegal in Europe it seems

😲 

Do you have any reference about that?

alecharp:
just from the issue on ionicons, there is https://www.theregister.com/2022/01/31/website_fine_google_fonts_gdpr/

CONTRIBUTING.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+If you want to simply run the project locally, you can run:
+
+[source,bash]
+----
+mvn package -Dmaven.test.skip
+docker compose up
+----
+
+NOTE: the environment variable required must be set in a file named `.env`.
```
lemeurherve, line 67:
```suggestion
NOTE: the required environment variables must be set in your local environment or in a file named `.env`.
```

lemeurherve (approved):
Some questions, LGTM otherwise.

pom.xml
```diff
@@ -22,7 +22,7 @@
   ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
   ~ SOFTWARE.
   -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
```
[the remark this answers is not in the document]
alecharp: First, I'm surprise it wasn't before.
The http is redirected to the https so, I guess it's better to use the https anyway.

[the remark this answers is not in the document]
alecharp: It this a good enough explanation so we can merge this?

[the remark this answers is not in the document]
lemeurherve: I still don't know why not all of them are in HTTPS, shouldn't be the case?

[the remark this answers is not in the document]
alecharp: - `xmlns="http://maven.apache.org/POM/4.0.0"` is the declaration of the default namespace
- `xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"` is the declaration of the `xsi` namespace
- `xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd"` is a way to say "the namespace $1 has a schema which can be retrieved at $2".

So the first 1 is just a placeholder, 2 could be https. For 3, the first part needs to match 1, the second can be https.

As the project is using `tidy-maven-plugin` to keep the pom ordered, this is in the end replaced by https://github.com/mojohaus/tidy-maven-plugin/blob/ba4716eb82177693dde1124d6bfeee3d1a6f6e15/src/main/java/org/codehaus/mojo/tidy/task/EnsureSingleLineProjectStartTag.java#L31-L33

lemeurherve (approved):

PR #98, opened 2022-10-27
Plugins can be removed from update-center

### LINKED ISSUE #97: Plugins can be removed from the list in the update-center
When a plugin is deprecated, it is removed from the `plugins` list in the update-center. 
This creates a problem for probes using this list and not testing that possibility (like the `UpForAdoptionProbe`).

alecharp:
When a plugin is deprecated, it is removed from the plugins list in the UC.

Resolves #97.

### DIFF
--- a/src/main/java/io/jenkins/pluginhealth/scoring/probes/UpForAdoptionProbe.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/probes/UpForAdoptionProbe.java
@@ -28,6 +28,8 @@ import io.jenkins.pluginhealth.scoring.model.Plugin;
 import io.jenkins.pluginhealth.scoring.model.ProbeResult;
 import io.jenkins.pluginhealth.scoring.model.updatecenter.UpdateCenter;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.core.annotation.Order;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/jenkins/pluginhealth/scoring/probes/UpForAdoptionProbe.java
```diff
@@ -37,15 +39,20 @@
...
     public static final int ORDER = DeprecatedPluginProbe.ORDER + 1;
 
     @Override
     public ProbeResult doApply(Plugin plugin, ProbeContext context) {
         final UpdateCenter updateCenter = context.getUpdateCenter();
-        if (updateCenter.plugins().get(plugin.getName()).labels().contains("adopt-this-plugin")) {
+        final io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin pl = updateCenter.plugins().get(plugin.getName());
+        if (pl == null) {
+            LOGGER.info("Couldn't not find {} in update-center", plugin.getName());
+            return ProbeResult.failure(key(), "Plugin is not in the update-center");
```
lemeurherve, line 51:
```suggestion
            return ProbeResult.failure(key(), "This plugin is not in the update-center");
```
(nit, to be similar to the other messages)

lemeurherve (approved):

src/main/java/io/jenkins/pluginhealth/scoring/probes/UpForAdoptionProbe.java
```diff
@@ -37,15 +39,20 @@
...
 
     @Override
     public ProbeResult doApply(Plugin plugin, ProbeContext context) {
         final UpdateCenter updateCenter = context.getUpdateCenter();
-        if (updateCenter.plugins().get(plugin.getName()).labels().contains("adopt-this-plugin")) {
+        final io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin pl = updateCenter.plugins().get(plugin.getName());
+        if (pl == null) {
+            LOGGER.info("Couldn't not find {} in update-center", plugin.getName());
+            return ProbeResult.failure(key(), "This plugin is not in the update-center");
+        }
```
dheerajodha, line 52: Sorry for adding a question after this PR has been merged already, but if you have time, can you please answer my 2 questions?

1. The PR description says that "When a plugin is deprecated, it is removed from the plugins list in the UC". However, when I take an example of a deprecated plugin, say, `BlameSubversion`. I see it both on the "deprecations" list as well as the "plugins" list. Is this expected?

2. As per my understanding, the above `if` statement would never become true. Because all the `plugin` objects passed to this `doApply` method are coming from the probe engine, which instead is getting the plugin objects from the DB (via `pluginService.streamAll()`), and DB got these plugin objects from the Update Center's `plugins().values()` (via the `saveOrUpdate` method). So basically the origin of these plugins are same as `updateCenter.plugins()` in the above line, so the `get()` method would always fetch and return a plugin object, hence it will never be `null`. Can you please tell me what I'm missing here?

Thanks!

alecharp: > However, when I take an example of a deprecated plugin, say, BlameSubversion. I see it both on the "deprecations" list as well as the "plugins" list. Is this expected?

Because BlameSubersion is deprecated but has no ongoing security warnings. This is not the case for others (`xframium` for example) which lead me to discover this behavior.

> As per my understanding, the above if statement would never become true.

Incorrect. 2 weeks ago, `xframium` was in the update-center and could be installed. It was in the `plugins` list of the update-center, so It was in my DB. last week, with a security advisory, the plugins was deprecated with a security warning and removed from the `plugins` list of the update-center. Because it was in my DB, the probe was executed on it but the map of plugins of the update-center doesn't have it anymore. So, this is not `true`.

dheerajodha: This is very useful information. I just looked at the first plugin in the deprecations object and got confused.
I'm glad I asked the question and thanks for letting me know.

PR #102, opened 2022-10-31
Lists the raw results of probes

### LINKED ISSUE #96: Update plugin-health.jenkins.io/probes to include the total raw data of the probes
As it stands right now with the probes we have, of the 1881 plugins:

120 are up for adoption
178 have public security issues
32 are deprecated
442 are configured with Dependabot
71 are configured with JEP-229 (continuous delivery - looking deeper into this one, seems too low)
1156 have a Jenkinsfile

It would be nice to see this data on that webpage maybe in its own column.

dheerajodha:
Fixes #96

This PR adds a new view at `/results` which lists the probe results as seen below:
image


src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Transactional
+    public Map<String, Integer> getProbeResultsData() {
+        Map<String, Integer> probeResultsMap = new HashMap<>();
+
+        pluginService.streamAll().forEach(
+            plugin -> {
+                plugin.getDetails().values().forEach(probeResult -> {
+                    if (probeResult.status() == ResultStatus.SUCCESS) {
```
dheerajodha, line 44: While rendering the new view: for a given probe, I count its value towards its result only if its `status()`, for a given plugin, is a `SUCCESS`. And for each probe, there's a different definition for `SUCCESS`.

For example, as per the current code, for a given plugin, the `deprecation` probe is a `SUCCESS` if the **plugin is NOT deprecated**.  Due to this, the result of this probe in the new view is **1851**, meaning, 1851 plugins are not deprecated. And since this is opposite to what @jleon33 's [message](https://github.com/jenkins-infra/plugin-health-scoring/issues/96#issue-1424226722) captures (it says "32 are deprecated"), I have a question on this.

I guess I understand why we chose the current definition of `SUCCESS` for the `deprecation` probe, it's better to see the 1851 SUCCESS status than the 1851 FAILURE status in the DB.

**[Question]:** So to tackle this:
- should we change the current definition of `SUCCESS` for these kinds of probes? or
- should I write an `if-else` chain in place of this line, where I would match the `status()` of each probe with a custom `ResultStatus` value?

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
```diff
@@ -0,0 +1,54 @@
...
+        );
+
+        return modelAndView;
+    }
+
+    @Transactional
+    public Map<String, Integer> getProbeResultsData() {
+        Map<String, Integer> probeResultsMap = new HashMap<>();
+
+        pluginService.streamAll().forEach(
```
dheerajodha, line 41: To collect the probe results data, I'm relying on the plugins stored in the database by performing an additional query to the DB. I scan their `details` value and accordingly populate the `probeResultsMap`. Is this an acceptable approach?

An alternate approach could be to populate this map while we're running the probe engine, in this way we wouldn't need to query the DB twice (as done in the above-mentioned approach). But later I realized that this might not be an ideal approach because when we're not running our probe engine, the `probeResultsMap` map would be empty and so will our `/results` page. Plus, when we're running the probe engine, this map will get populated, but this will very happen very slowly as the probe engine takes some time in processing each probe.

wdyt?

alecharp:
This is a good start I think. Thanks @dheerajodha. What I think could improve the outcome:
 1. having the total number of plugins known in the result column (like `1800 / 1881` for the `last-commit-date`)
 2. I wonder if there is a `Collectors` implementation which would simplify the `if` in the controller
 3. I don't think having the transactional (which is good) method in the controller if ok. It might be better in the PluginService class.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
alecharp: > should we change the current definition of SUCCESS for these kinds of probes?

I'm really against this. It is a "bad" thing to see a plugin deprecated, so it's a failure.

Maybe the description of the probe needs to refer to the expected behavior of the probe more clearly.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
alecharp: > To collect the probe results data, I'm relying on the plugins stored in the database by performing an additional query to the DB.

I don't understand this. Could you please detail what you mean?

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
dheerajodha: > I'm really against this. It is a "bad" thing to see a plugin deprecated, so it's a failure.

:+1: 

> Maybe the description of the probe needs to refer to the expected behavior of the probe more clearly.

I think that is a good idea, it makes it easier to know the success criteria for a probe. Because currently if I want to know the success criteria for a probe, I look at the code and look for the `message` String for `ProbeResult.success()` (which is not difficult to find as well).

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
dheerajodha: > I don't understand this. Could you please detail what you mean?

Yes, sorry I should've mentioned that I was referring to the `pluginService.streamAll()` (used in line 41 above) method which does an "additional query" to the DB. So I thought about maybe there's a way where we'd not even need to query the DB, and then I started thinking about the "alternate approach" as I mentioned above. But I can see now why this alternate approach is a bad idea.


> Which means we wouldn't need to have the DB and we would have everything in memory. I don't think this is a good idea.
> 

:smile: true. I need to keep in mind that having everything in memory is not good, and there's a reason we have concept of DB.


> To get the number to Jake, I ran this SQL query:
> 
> ```
> SELECT
>     count(id) as total,
>     (SELECT count(id) FROM plugins WHERE scm IS NOT NULL AND details -> 'up-for-adoption' ->> 'status' = 'FAILURE') as upForAdoption,
>     (SELECT count(id) FROM plugins WHERE scm IS NOT NULL AND details -> 'security' ->> 'status' = 'FAILURE') as security,
>     (SELECT count(id) FROM plugins WHERE scm IS NOT NULL AND details -> 'deprecation' ->> 'status' = 'FAILURE') as deprecated,
>     (SELECT count(id) FROM plugins WHERE scm IS NOT NULL AND details -> 'dependabot' ->> 'status' = 'SUCCESS') as dependabot,
>     (SELECT count(id) FROM plugins WHERE scm IS NOT NULL AND details -> 'jep-229' ->> 'status' = 'SUCCESS' ) as jep229,
>     (SELECT count(id) FROM plugins WHERE scm IS NOT NULL AND details -> 'jenkinsfile' ->> 'status' = 'SUCCESS') as jenkinsfile
> FROM plugins
> WHERE scm IS NOT NULL;
> ```
> 
> Maybe you can map that to a (or multiple) method signature in `PluginRepository` and have a method in `PluginService` to return that map.

Wow, this is the first time I saw an SQL query being run on a JSON object. Thanks for sharing the SQL code. I'll think about this now.

src/main/resources/templates/probes/results.html
```diff
@@ -0,0 +1,28 @@
...
+        <thead>
+        <tr>
+            <th>ID</th>
+            <th>Result</th>
+        </tr>
+        </thead>
+        <tbody>
+        <tr data-th-each="probeResult: ${probeResults}">
+            <td><code data-th-text="${probeResult.key}"></code></td>
+            <td data-th-text="${probeResult.value}"></td>
```
dheerajodha, line 21: > 1. having the total number of plugins known in the result column (like `1800 / 1881` for the `last-commit-date`)

Which one do we like more? This:
* image

OR this?

* image

alecharp: The first one.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
```diff
@@ -0,0 +1,50 @@
...
+        Map<String, Integer> probeResultsMap;
+
+        probeResultsMap = pluginService.streamAll()
+            .map(plugin ->
+                plugin.getDetails().values().stream().filter(probeResult -> probeResult.status() == ResultStatus.SUCCESS)
+                .collect(Collectors.toMap(probeResult -> probeResult.id(), probeResult -> 1)))
+            .flatMap(m -> m.entrySet().stream())
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (currentProbeData, latestProbeData) -> currentProbeData + 1));
+
+        probeResultsMap.put("size", (int) pluginService.streamAll().count());
```
alecharp, line 46: Might be more efficient to have a `count` method on the service.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public Map<String, Integer> getProbeResultsData() {
+        Map<String, Integer> probeResultsMap;
+
+        probeResultsMap = pluginService.streamAll()
+            .map(plugin ->
+                plugin.getDetails().values().stream().filter(probeResult -> probeResult.status() == ResultStatus.SUCCESS)
+                .collect(Collectors.toMap(probeResult -> probeResult.id(), probeResult -> 1)))
+            .flatMap(m -> m.entrySet().stream())
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (currentProbeData, latestProbeData) -> currentProbeData + 1));
```
alecharp, line 44: I don't understand the need to the two collectors. Couldn't it be simplified? Something like 

```suggestion
            .flatMap(plugin -> plugin.getDetails().values().stream())
            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (currentProbeData, latestProbeData) -> currentProbeData + 1));
```

alecharp (changes requested):

dheerajodha:
> I am just looking in more detail at some of the stats that @dheerajodha is pulling, and they arent making any sense? Specifically, deprecation, up for adoption, and size(?).

Sorry for this mess Jake, it's my fault :smile:

To answer your question,
- "deprecation", "up-for-adoption", and "security" stats are not making sense because I blindly collected raw data for each probe based on the number of plugins which has a **ResultStatus == "SUCCESS"**.
    - This is not right because for a probe, say, "deprecation", we want to see the stats for plugins that are deprecated. But in the code of this probe, we've given **ResultStatus = "SUCCESS"** for those plugins which are **NOT** deprecated. So, my code was capturing (and displaying) the opposite of the expected data.
    - Adrien suggested I create a custom mapping between a "probe" and "status" (for which we want to capture the data), so I just committed a fix for this based on his suggestions.
    - My code still needs to be reviewed, but you can [see this mapping here](https://github.com/jenkins-infra/plugin-health-scoring/pull/102/files#diff-858a187ff159e654e8ac5b13d631020aeac98f58a3772eba24c489e6e4d1d710R66-R83). Notice that we've now passed "FAILURE" status to capture the expected data for these 3 probes.

- regarding the "size": it was my dumb hack to display the total count of plugins on the UI (as you can see "/ 1881" in each row of the "Result" column).

I've fixed them now (I think), and the UI looks like this:
image

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
dheerajodha: Thank you for catching this, I've added the `getPluginsCount()` method to the service.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
dheerajodha: Thanks! this makes much more sense.
Based on the SQL query you shared above, I've further refactored this method to incorporate the changes related to the custom mapping between probes and their status, to capture expected raw data.

alecharp:
> Also, we may want to get creative with the last commit stat, as the raw count doesn't really do us any good. Maybe show an average?

those are row data. Getting the last commit date is helpful while scoring but just its value has no meaning. I don't think we need to show that here.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Controller
+@RequestMapping(path = "/results")
+public class ProbeResultController {
+
+    private final List<Probe> probes;
+    private final PluginService pluginService;
+
+    public ProbeResultController(List<Probe> probes, PluginService pluginService) {
+        this.probes = probes;
```
alecharp, line 23:
```suggestion
        this.probes = List.of(probes);
```

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -56,4 +56,33 @@ public Plugin saveOrUpdate(Plugin plugin) {
...
+            case "security":
+                return pluginRepository.getProbeRawData("security", "FAILURE");
+            case "deprecation":
+                return pluginRepository.getProbeRawData("deprecation", "FAILURE");
+            case "dependabot":
+                return pluginRepository.getProbeRawData("dependabot", "SUCCESS");
+            case "jep-229":
+                return pluginRepository.getProbeRawData("jep-229", "SUCCESS");
+            case "jenkinsfile":
+                return pluginRepository.getProbeRawData("jenkinsfile", "SUCCESS");
```
alecharp, line 83:
```suggestion
            case "scm":
            case "last-commit-date":
            case "dependabot":
            case "jep-229":
            case "jenkinsfile":
                return pluginRepository.getProbeRawData(probeID, "SUCCESS");
            case "up-for-adoption":
            case "security":
            case "deprecation":
                return pluginRepository.getProbeRawData(probeID, "FAILURE");
```

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -56,4 +56,27 @@ public Plugin saveOrUpdate(Plugin plugin) {
...
+            case "dependabot":
+            case "jep-229":
+            case "jenkinsfile":
+                return pluginRepository.getProbeRawData(probeID, "SUCCESS");
+            case "up-for-adoption":
+            case "security":
+            case "deprecation":
+                return pluginRepository.getProbeRawData(probeID, "FAILURE");
+            default:
+                return 0;
```
alecharp, line 79: This means that when we will be adding new probes in the future, we will always have to add it here. I don't think this is practical.

```suggestion
        switch (probeID) {
            case "up-for-adoption":
            case "security":
            case "deprecation":
                return pluginRepository.getProbeRawData(probeID, "FAILURE");
            default:
                return pluginRepository.getProbeRawData(probeID, "SUCCESS");
```

src/main/resources/templates/probes/results.html
```diff
@@ -0,0 +1,28 @@
...
+        <thead>
+        <tr>
+            <th>ID</th>
+            <th>Result</th>
+        </tr>
+        </thead>
+        <tbody>
+        <tr data-th-each="probeResult: ${probeResults}">
+            <td><code data-th-text="${probeResult.key}"></code></td>
+            <td data-th-text="${probeResult.value} + ' / ' + ${@pluginService.getPluginsCount()}"></td>
```
alecharp, line 21: Please, don't access DB services on the UI.

alecharp (changes requested):

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
dheerajodha: Done, thanks

src/main/resources/templates/probes/results.html
dheerajodha: Now accessing via the `modelAndView`'s new `attributeName`.

src/main/java/io/jenkins/pluginhealth/scoring/repository/PluginRepository.java
```diff
@@ -29,9 +29,15 @@
...
 import org.springframework.stereotype.Repository;
 
 @Repository
 public interface PluginRepository extends JpaRepository<Plugin, Long> {
     Optional<Plugin> findByName(String name);
+
+    @Query(
+        value = "SELECT count(id) FROM plugins WHERE scm IS NOT NULL AND details -> ?1 ->> 'status' = ?2",
+        nativeQuery = true)
+    int getProbeRawData(String probeID, String status);
```
alecharp, line 42: This to match the `count` method return type.

```suggestion
        nativeQuery = true
    )
    long getProbeRawData(String probeID, String status);
```

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
```diff
@@ -0,0 +1,70 @@
...
+import io.jenkins.pluginhealth.scoring.probes.Probe;
+import io.jenkins.pluginhealth.scoring.service.PluginService;
+
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.servlet.ModelAndView;
+
+@Controller
+@RequestMapping(path = "/results")
```
alecharp, line 40: General question: why not have this behind `/probes/results` and have the code in the `ProbesController`?

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
```diff
@@ -0,0 +1,70 @@
...
+            "pluginsCount",
+            pluginService.getPluginsCount()
+        );
+
+        return modelAndView;
+    }
+
+    public Map<String, Integer> getProbeResultsData() {
+        return probes.stream()
+            .collect(Collectors.toMap(probe -> probe.key(), probe -> pluginService.getProbeRawData(probe.key())));
```
alecharp, line 68:
```suggestion
            .collect(Collectors.toMap(Probe::key, probe -> pluginService.getProbeRawData(probe.key())));
```

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
```diff
@@ -0,0 +1,70 @@
...
+            getProbeResultsData()
+        ).addObject(
+            "pluginsCount",
+            pluginService.getPluginsCount()
+        );
+
+        return modelAndView;
+    }
+
+    public Map<String, Integer> getProbeResultsData() {
```
alecharp, line 66: To match return type change

```suggestion
    public Map<String, Long> getProbeResultsData() {
```

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -56,4 +56,21 @@ public Plugin saveOrUpdate(Plugin plugin) {
...
         return pluginRepository.findAll().stream();
     }
+
+    @Transactional
+    public int getPluginsCount() {
+        return pluginRepository.findAll().size();
+    }
+
+    @Transactional
+    public int getProbeRawData(String probeID) {
```
alecharp, line 66: To match return type change

```suggestion
    public long getProbeRawData(String probeID) {
```

src/main/java/io/jenkins/pluginhealth/scoring/service/PluginService.java
```diff
@@ -56,4 +56,21 @@ public Plugin saveOrUpdate(Plugin plugin) {
...
+        switch (probeID) {
+            case "up-for-adoption":
+            case "security":
+            case "deprecation":
+                return pluginRepository.getProbeRawData(probeID, "FAILURE");
+            default:
+                return pluginRepository.getProbeRawData(probeID, "SUCCESS");
+        }
```
alecharp, line 74: Found out this syntax.

```suggestion
        return switch (probeID) {
            case "up-for-adoption", "security", "deprecation" ->
                pluginRepository.getProbeRawData(probeID, "FAILURE");
            default ->
                pluginRepository.getProbeRawData(probeID, "SUCCESS");
        };
```

src/main/resources/templates/layouts/main.html
```diff
@@ -20,6 +20,9 @@
     <a href="" data-th-href="@{/}" class="page-navbar__item page-navbar__item-selected">
         <ion-icon name="home"></ion-icon>
     </a>
+    <a href="" data-th-href="@{/results}" class="page-navbar__item page-navbar__item-selected">
```
alecharp, line 23: the `page-navbar__item-selected` should only be applied when the user is on that specific page. 
`home` button has it because no other page was available.

alecharp (changes requested):

src/main/resources/templates/layouts/main.html
dheerajodha: After 2hrs I was able to do it (I think), thanks for the correction!

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbeResultController.java
dheerajodha: I've been trying to do this and the `probes/results` view gets built successfully with expected results but I broke the navbar on the left. The 2 icons are not visible when:
- we're on the Probe Results page (via `/probes/results`),
- and while being on the `probes/results` page, when we delete `results` from this URL and hit ENTER.

Any hints please?

alecharp: I think it's because of the ionicons library which is not able to grab the svg file correctly.

dheerajodha: Okay I'll check, but the icons are visible in one view and not in the other, makes me wonder how are these icons dependent on the mapping. Before my latest commit, everything works as expected. But I didn't touch any <ion-icons> tags or *.svg files in this latest commit...

alecharp: > Okay I'll check, but the icons are visible in one view and not in the other, makes me wonder how are these icons dependent on the mapping

if you look at your "network" tab in the developer tools of the web browser you are using, you'll see that the icons are fetched from different paths depending if you are on `/probes` or `/probes/results`, which shouldn't be the case.

dheerajodha: Thanks!! On the "network" tab I noticed that the `probes/results` view was appending "probes" before the icon path, making it incorrect. So, I explicitly specified the icon path starting with `/` so that thymeleaf doesn't add the prefix "probes" to it. And this "probes" prefix was coming from `GetMapping("probes")`.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbesController.java
```diff
@@ -59,6 +63,30 @@ public ModelAndView list() {
...
+
+    public Map<String, Long> getProbeResultsData() {
+        Map<String, Long> probeResultsMap = probes.stream()
+            .collect(Collectors.toMap(Probe::key, probe -> pluginService.getProbeRawData(probe.key())));
+
+        probeResultsMap.remove("last-commit-date");
+
+        return probeResultsMap;
+    }
+
```
alecharp, line 89: Normally, we don't mix view layer with service layer. 
I'll be more comfortable having this method in a `ProbeService` which would be injected in this controller.

alecharp (changes requested):

src/main/resources/templates/layouts/main.html
```diff
@@ -17,8 +17,11 @@
 <body id="app">
 
 <nav id="page-navbar" class="navbar">
-    <a href="" data-th-href="@{/}" class="page-navbar__item page-navbar__item-selected">
-        <ion-icon name="home"></ion-icon>
+    <a data-th-href="@{/}" class="page-navbar__item" th:classappend="${#request.requestURI == '/probes' ? 'page-navbar__item-selected' : ''}">
+        <ion-icon name="home" data-th-src="@{/svg/home.svg}"></ion-icon>
```
alecharp, line 21: this is definitively not the solution.

alecharp (commented):
This is not the fix for the svg images. Let's put a ticket in GH and work on this in a different pull request.

src/main/java/io/jenkins/pluginhealth/scoring/http/ProbesController.java
dheerajodha: Done. I also moved  `getProbesRawResultsFromDatabase` (previously called `getProbeRawData`) from `PluginService` to `ProbeService`. And changed the method names to try to make it a little clear what they do.

src/main/resources/templates/layouts/main.html
dheerajodha: Hmm, why is it "definitively not" the solution? I guess it's because I've hardcoded the path, right?

dheerajodha: Sure, I'd revert this.

src/main/resources/templates/layouts/main.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" xmlns:th="http://www.w3.org/1999/xhtml">
```
alecharp, line 2: I missed that. It's incorrect.
```suggestion
<html lang="en">
```

src/main/resources/templates/layouts/main.html
```diff
@@ -17,9 +17,12 @@
 <body id="app">
 
 <nav id="page-navbar" class="navbar">
-    <a href="" data-th-href="@{/}" class="page-navbar__item page-navbar__item-selected">
+    <a href="" data-th-href="@{/}" class="page-navbar__item" th:classappend="${#request.requestURI == '/probes' ? 'page-navbar__item-selected' : ''}">
```
alecharp, line 20: Never ever use `==` to compare strings.

```suggestion
    <a href="" data-th-href="@{/}" class="page-navbar__item" data-th-classappend="${#request.requestURI.equals('/probes') ? 'page-navbar__item-selected' : ''}">
```

dheerajodha: Sorry for that, I'll keep that in mind.

src/main/java/io/jenkins/pluginhealth/scoring/service/ProbeService.java
```diff
@@ -0,0 +1,44 @@
+package io.jenkins.pluginhealth.scoring.service;
```
alecharp, line 1: Could you add the license here please?

alecharp (changes requested):

src/main/java/io/jenkins/pluginhealth/scoring/service/ProbeService.java
dheerajodha: Done, thanks

src/main/resources/templates/layouts/main.html
dheerajodha: Are you able to see the icons on `probes/results` after this commit: https://github.com/jenkins-infra/plugin-health-scoring/pull/102/commits/4fafd29f4e586193ccc1dc9f9ff4bd4e5e76d7be? I'm still not seeing them after I pulled the latest changes, am I missing something?

alecharp (approved):

PR #111, opened 2022-11-16
Validates plugin documentation migration

### LINKED ISSUE #58: New probe: is the plugin documentation located on GitHub?

alecharp:
Resolves #58 

Followup of #71


alecharp:
At this moment, the Maven project properties are not expanded to express the real value of the url. 
This needs to be addressed to be able to move forward.

impact-maker:
> At this moment, the Maven project properties are not expanded to express the real value of the url. This needs to be addressed to be able to move forward.

hi @alecharp  Could you elaborate this blocker. This might help me collaborate with you in order to close this issue.

alecharp:
At this time, the implementation done on this PR will have to much rework to do because of #206. I'm closing this. The requirement to get a probe about the documentation is still very much present.

PR #112, opened 2022-11-16
Improve scoring engine testing

alecharp:
Changeset coming from suggestions on #105 from @dheerajodha.


dheerajodha:
Thank you for your time on this PR.
Also, I think [this line](https://github.com/jenkins-infra/plugin-health-scoring/blob/main/src/test/java/io/jenkins/pluginhealth/scoring/scores/ScoreEngineTest.java#L109) is duplicating line 107 above it.

PR #116, opened 2022-11-17
Displays application version in production

### LINKED ISSUE #63: Display application version in production
@lemeurherve suggested that the application version should be visible when deployed in production. This way we can track what is deployed.

This is done in
https://github.com/jenkins-infra/plugin-health-scoring/blob/6f9c5d2363df9599117e310f88f4461f61249ba5/src/main/resources/templates/layouts/main.html#L26-L27

but limited to the `dev` profile. We _just_ need to remove the `data-th-if` and put the `commitId` in a title.

alecharp:
Resolves #63.

### DIFF
--- a/src/main/resources/templates/layouts/main.html
+++ b/src/main/resources/templates/layouts/main.html
@@ -35,10 +35,7 @@
 </main>
 <footer class="page-footer">
     <span class="grow"></span>
-    <span data-th-if="${#arrays.contains(@environment.getActiveProfiles(), 'dev')}"
-          class="app--version"
-          data-th-text="${@buildProperties.version} + ' - ' + ${@gitProperties.shortCommitId}">
-    </span>
+    <span class="app--version" data-th-text="${@buildProperties.version}"></span>
 </footer>
 
 </body>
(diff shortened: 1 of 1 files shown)


lemeurherve (commented):
Is it possible to get the commit id as alt text visible on hoover?
LGTM otherwise, thanks for taking care of this issue, completely forgot about it.

alecharp:
In the end, having the commit id in a tooltip way was not really usable to me. So I'm displaying it all the time.

src/main/resources/templates/layouts/main.html
```diff
@@ -35,10 +35,8 @@
 </main>
 <footer class="page-footer">
     <span class="grow"></span>
-    <span data-th-if="${#arrays.contains(@environment.getActiveProfiles(), 'dev')}"
-          class="app--version"
-          data-th-text="${@buildProperties.version} + ' - ' + ${@gitProperties.shortCommitId}">
-    </span>
+    <span class="app--version"
+          data-th-text="${@buildProperties.version} + ' - ' + ${@gitProperties.shortCommitId}"></span>
```
lemeurherve, line 39:
```suggestion
          data-th-text="${@buildProperties.version} + ' (' + ${@gitProperties.shortCommitId} + ')'"></span>
```

lemeurherve (approved):
Nit, I would put the commit id in parenthesis to de-emphasize it, LGTM in both case.

PR #117, opened 2022-11-17
Fixes probes expected context

### LINKED ISSUE #115: Suspended plugins are not detected
Plugins distribution can be suspended. When this occurs, the plugin completely disappears from the update-center.

To handle this situation we could: 
 - delete the the plugin from the DB
 - have a marker on the plugin entry and ignore it in the probe-engine

Either way, we need to be able to detect those plugins and handle them correctly in the application. Currently, this is throwing exceptions in the logs. 

See https://github.com/jenkins-infra/update-center2/blob/master/resources/artifact-ignores.properties

alecharp:
Resolves #115 

For some probes to have a chance to be successful, some previously executed probes result can be tested. This could prevent unexpected behavior of the code.

### DIFF
--- a/src/main/java/io/jenkins/pluginhealth/scoring/probes/Probe.java
+++ b/src/main/java/io/jenkins/pluginhealth/scoring/probes/Probe.java
@@ -36,7 +36,7 @@ public abstract class Probe {
      * Starts the analyze on a plugin.
      * Should only be called by the {@link ProbeEngine#run()} method.
      *
-     * @param plugin the plugin on which to perform the analyze
+     * @param plugin  the plugin on which to perform the analyze
      * @param context holder of information passed across the probes executed on a single plugin
      * @return the result of the analyze in a {@link ProbeResult}
      */
@@ -47,7 +47,7 @@ public abstract class Probe {
     /**
      * Perform the analyze on a plugin
      *
-     * @param plugin the plugin on which the analyze is done
+     * @param plugin  the plugin on which the analyze is done
      * @param context holder of information passed across the probes executed on a single plugin
      * @return a ProbeResult representing the result of the analyze
      */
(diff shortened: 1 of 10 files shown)


alecharp:
I'm not sure this is true fix for the errors which led to create #115 but at least it will help reduce the noise in the logs.

src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -65,7 +66,10 @@ public SCMLinkValidationProbe(HttpClient httpClient, GithubConfiguration githubC
     }
 
     @Override
-    public ProbeResult doApply(Plugin plugin, ProbeContext context) {
+    public ProbeResult doApply(Plugin plugin, ProbeContext context) {        final ProbeResult deprecatedProbeResult = plugin.getDetails().get(DeprecatedPluginProbe.KEY);
+        if (deprecatedProbeResult != null && deprecatedProbeResult.status().equals(ResultStatus.FAILURE)) {
```
lemeurherve, line 70:
```suggestion
    public ProbeResult doApply(Plugin plugin, ProbeContext context) {
        final ProbeResult deprecatedProbeResult = plugin.getDetails().get(DeprecatedPluginProbe.KEY);
        if (deprecatedProbeResult != null && deprecatedProbeResult.status().equals(ResultStatus.FAILURE)) {
```
Missing line return.

lemeurherve (changes requested):
A missing line return, otherwise LGTM

src/main/java/io/jenkins/pluginhealth/scoring/probes/Probe.java
```diff
@@ -36,7 +36,7 @@ public abstract class Probe {
      * Starts the analyze on a plugin.
      * Should only be called by the {@link ProbeEngine#run()} method.
      *
-     * @param plugin the plugin on which to perform the analyze
+     * @param plugin  the plugin on which to perform the analyze
```
lemeurherve, line 39: I'd remove this additional space (same below)

lemeurherve (approved):
LGTM

src/main/java/io/jenkins/pluginhealth/scoring/probes/Probe.java
alecharp: It's the formatting so that the definitions start on the same column for the parameters.

PR #133, opened 2022-11-25
Provides GitHub health indication

### LINKED ISSUE #130: Using GitHub Application in production to reach GitHub API
Currently, the application is using a GitHub OAuth token to access the API. 
This method is fine but have limitations: numbers of call possible per hours but also the token expired. This is not great for production.

alecharp:
This will help diagnose what is happening with the application and understand why some probes are not executed.


alecharp:
The current implementation would mean that the application keeps being deployment by Kubernetes if the configuration to GitHub is incorrect. This is not what we want.

alecharp:
With https://github.com/jenkins-infra/helm-charts/pull/328, we can specify that the GitHub connection is down without  triggering a new deployment to happen from Kubernetes

PR #141, opened 2022-12-02
Fixes meaning of CD

alecharp:


src/main/java/io/jenkins/pluginhealth/scoring/probes/ContinuousDeliveryProbe.java
```diff
@@ -75,6 +75,6 @@ public String key() {
 
     @Override
     public String getDescription() {
-        return "Checks if JEP-229 (Continuous Deployment) has been activated on the plugin";
+        return "Checks if JEP-229 (Continuous Delivery) has been activated on the plugin";
```
olamy, line 78: interesting to take this into account...
some contributors do not activate this because it means depending on a concurrent of Jenkins (e.g Github Actions)
does it mean we really consider a plugin more healthy because it promotes the use of Github Actions?

alecharp: The problem is that some plugins are not released enough and JEP-229 is helping with that. 
If the JEP was implemented using Jenkins it would be great, but for now it's not possible.

Issue #155, opened 2022-12-07
Plugin uniq key is not respected


alecharp:
When updating the plugin list locally, this exception was thrown

```
org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "plugins_pkey"
  Detail: Key (id)=(502) already exists.
```

This needs to be addressed urgently.

alecharp:
I was able to reproduce (outside of a test) the problem, but now I cannot, even less with a test.

alecharp:
This issue as not been reproduced for quite some time now. I'm closing it for now.

PR #168, opened 2023-01-10
Improves plugins update test suite

alecharp:
First, was an attempt to work on #155 but again, the bug was not reproducible with tests and it completely stopped from happening locally. 
So this is just to had more tests.


alecharp:
And the problem is not showing in my local instance anymore.

PR #159, opened 2022-12-13
Use GitHub App

alecharp:
Resolves #130.

PR #199, opened 2023-01-29
docs: fixed the contributing documentation link

Jagrutiti:
For some reason the GitHub wasn't able to process the AsciiDoc syntax to include code from another.

Updated the link of the `env.examples`  to be referenced as hyperlink.

Reference: https://docs.asciidoctor.org/asciidoc/latest/directives/include/#include-nonasciidoc


Jagrutiti:
Hey @alecharp 

The earlier syntax wasn't rendered. It printed the code as it is: `include::.env.example[]`. As a reader, this did not make sense to me. 

My current syntax will create a hyperlink to the file. In this case, the user can view the file themselves

alecharp:
> The earlier syntax wasn't rendered.

In GitHub? That is normal, GitHub is not rendering AsciiDoc file correctly. 

Screenshot 2023-01-30 at 14 08 26

This is how it is rendered normally.

> My current syntax will create a hyperlink to the file. In this case, the user can view the file themselves

No, your current syntax is not correct and is not creating the link your are expecting.

Jagrutiti:
> In GitHub? That is normal, GitHub is not rendering AsciiDoc file correctly.

For me the file is rendered this way:

image

The syntax is rendered in Intellij but not in GitHub.

Hence the PR. 

>Please install the AsciiDoc plugin on your IDE to see what you are proposing.

Do you still want me to continue with the PR?

alecharp:
> The syntax is rendered in Intellij but not in GitHub.

Because GitHub is not rendering AsciiDoc files correctly. But it doesn't mean the syntax of the file is broken.

> Do you still want me to continue with the PR?

You can, but the content of the PR is not correct so I won't merge it in this state.

You can add a few words to the original content so say "you can find an example of the variables in `.env.example`" and keep the `include` for example.

But as is, the content is not correct and I won't accept it.

lemeurherve:
The code is hosted on GitHub, people will look at the README there most of the time. I find it unfortunate having a broken rendering by default, especially for a section needed for running the project locally.

alecharp:
But for the contributing guide, shouldn't it be read from the IDE? Even more for details about how to run the application locally? As MD has no `include` equivalent, it would mean to duplicate the element which I find source of issues when you change them (like when I did implement #159)

alecharp:
@Jagrutiti

> Do you still want me to continue with the PR?

No, the documentation seems to be misleading to you, so it might to others. Fixing it is the way to go. I just don't feel you proposed the correct solution so far. Please see https://github.com/jenkins-infra/plugin-health-scoring/pull/199#issuecomment-1408849997 as I suggest something to move forward.

Jagrutiti:
Hey @alecharp 

What do you think of adding a GitHub link?

It is easier than searching for a file in project directory. I have made this change in my recent commit.

If not I will follow your last comment.

alecharp:
@Jagrutiti please don't include your `.idea/` files, they should not be in the git tree.

I would like you keep the `include` block and use a proper link to the file. You also included a typo in your change.

CONTRIBUTING.adoc
```diff
@@ -31,12 +31,9 @@ mvn package -Dmaven.test.skip
 There are two ways to run the project locally.
 Either way, the application requires some environment variables to be configured and a GitHub App.
 
-* Environment variables:
+* Environment variables: To create your own environment variables refer to:g
+https://github.com/Jagrutiti/plugin-health-scoring/blob/main/.env.example[.env.example]
```
alecharp, line 35:
```suggestion
* Environment variables: to create your own environment variables refer to link:.env.example[].
```

alecharp:
@Jagrutiti Please revert the deletion of the `include` and revert the addition of the files from `.idea`. Then I'll merge this PR.

Jagrutiti:
Hey @alecharp

I have made the change just like you wanted.

alecharp:
Please see that #203 is targeting a larger refactoring of the documentation. I included your suggestion there directly.
I'd appreciate if you could review that pull request, to make sure that the new documentation is clear.

PR #206, opened 2023-02-07
Split the application into modules

alecharp:
Currently, the application code is in one big module. 

WARNING: This is not to create a micro service architecture.

Having multiple modules can help understand the place of each component of the project. 
This can also help in case someone wants to extend the project behavior, even if submitting pull request here is the preferable way.


.dockerignore
```diff
@@ -1,10 +1,17 @@
 pom.xml
```
lemeurherve, line 1: Thinking out loud, wouldn't it be easier to ignore everything then adding only what's needed instead of listing each folder one per one?

pom.xml
```diff
@@ -26,10 +26,11 @@
 	<modelVersion>4.0.0</modelVersion>
```
lemeurherve, line 26: Global question about the different POMs: shouldn't they all use spaces instead of tabs as indents? Or the inverse?

war/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM eclipse-temurin:17.0.4.1_1-jre as builder
```
lemeurherve, line 1: Nit: I would use an ARG for the version, easier to keep up to date with an updatecli manifest.

war/src/main/resources/banner.txt
```diff
@@ -0,0 +1,9 @@
+
```
lemeurherve, line 1:
```suggestion
```
(nit)

lemeurherve (approved):
Great job! 👏

.dockerignore
alecharp: something like

```
*

!war/src/main/docker/Dockerfile
!war/src/main/docker/docker-entrypoint.sh
!war/target/plugin-health-scoring.jar
```

pom.xml
alecharp: I set that in `.editorconfig`. `pom.xml` files are using tab. We could change that.

war/src/main/docker/Dockerfile
alecharp: can the arg be after the declaration of the base image? 🤔

.dockerignore
lemeurherve: Something like that yes, easier to maintain I think.

alecharp:
I'm waiting for #198 #211 to be merged before merging this, so I can deal with the conflicts here.

PR #208, opened 2023-02-08
Enable Remote Development with Codespaces

MondalSayantan:
A codespace is a development environment that's hosted in the cloud. Each codespace you create is hosted by GitHub in a Docker container, running on a virtual machine. One can choose from a selection of virtual machine types, from 2 cores, 8 GB RAM, and 32 GB storage, up to 32 cores, 64 GB RAM, and 128 GB storage.

This PR adds the configuration to install and use correct versions of Java and Maven required for this project.

Note: Not to be confused with Gitpod

PR #209, opened 2023-02-08
Setting Gitpod Development Environment

MondalSayantan:
This PR allows to spin up a Gitpod development environment fast with the tools required for running this project.

1. Builds a Dockerfile which has
    - `gitpod/workspace-postgres` as the base image
    - Sets Java `17.0.3-ms` as the default
    - Node version `18.14.0`
    - Maven version `3.9.0`
    - Docker, PostgreSQL and Yarn come preinstalled with the image
  
2. Runs `mvn package -Dmaven.test.skip` on start


MondalSayantan:
I think I have added a file by mistake and messed something up

MondalSayantan:
This should now work just fine

alecharp:
So what was the problem with #208? Is gitpod fix now that we can use it?

MondalSayantan:
> So what was the problem with #208? Is gitpod fix now that we can use it?
> 
Before the gitpod.Dockerfile, I was facing these issues
image
image

I am able to now run it in Gitpod and run the command `mvn package -Dmaven.test.skip` without any errors.

> Looking at some documentation, it looks to me that the configuration is not complete nor doing what we would need.

I am really new in this project and I am still tailoring the development environment to my needs so I might make some mistakes. I have configured it so that it can run the `mvn` command without the issue of having an old version of Java. It would be really helpful you could point me to the direction of what else is needed so that I can work on it.

MondalSayantan:
Short note on what I have done to setup the development environment on Gitpod

- Pulls the complete Gitpod Workspace Image
- Pulls the Postgres Image which comes with some utility scripts
- Sets the Node.js version to 18.14.0
- Runs `mvn package -Dmaven.test.skip` on start
- Tried and tested it on VS Code for Browser

Notes

1. The PostgeSQL database automatically starts and binds on port 5432. To use PostgreSQL using Docker, we have to use the utility script `pg_stop` to stop the database.
2. #208 is also for setting up a cloud-based development environment, but on GitHub Codespaces. This is PR in only for Gitpod

MondalSayantan:
Hey @alecharp 
I realize my commits and comments might be messy but I hope I am heading in the right direction of enabling newcomers (like me) to set up a development environment quickly.

.gitpod.Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+# Set up Java environment
+RUN bash -c '. /home/gitpod/.sdkman/bin/sdkman-init.sh && \
+    sdk install java 17.0.3-ms && \
+    sdk default java 17.0.3-ms'
+
+# Set up Node.js environment
+RUN bash -c 'VERSION="18.14.0" \
+    && source $HOME/.nvm/nvm.sh && nvm install $VERSION \
+    && nvm use $VERSION && nvm alias default $VERSION'
```
alecharp, line 17: Shouldn't we do the same for Maven to make sure of the version it is using?

MondalSayantan: Thank you @alecharp, I have resolved this by pushing another commit which will ensure the maven version to be 3.9.0

.gitpod.Dockerfile
```diff
@@ -0,0 +1,27 @@
+#Pull Gitpod Workspace Image
```
alecharp, line 1: I just noticed that but could you move the file to a `.gitpod` folder and name it only `Dockerfile`?

alecharp (changes requested):
Just to organize a bit the root folder of the project.

MondalSayantan:
I tried moving the `.gitpod.yml` file as well to the `.gitpod` folder to clear the top level directory, but Gitpod does not pick up the configuration from there. This is an open issue in Gitpod https://github.com/gitpod-io/gitpod/issues/729

I have moved the Docker file successfully.

alecharp:
Yes, sorry I wasn't clearer but I meant to keep the `.gitpod.yml` file at the root level (because this is required, as you discovered) and then move the `.gitpod.Dockerfile` to `.gitpod/Dockerfile` (and so renaming it).

.gitpod/Dockerfile
```diff
@@ -0,0 +1,27 @@
+#Pull Gitpod Workspace Image
+FROM gitpod/workspace-full
+
+#Pull Postgres Image
+FROM gitpod/workspace-postgres
```
alecharp, line 5: You are pulling the first image but aren't doing anything with it. 
Do you mean to base your image from the first or second one?

alecharp (changes requested):

.gitpod/Dockerfile
MondalSayantan: I am extending the Gitpod base image `gitpod/workspace-full`because it provides many tools which might be required including Docker support which I am quoting below from the documentation

`Docker support: If you use the gitpod/workspace-full image, you get Docker support built-in to your environment.`

I hope this clarifies my intention

MondalSayantan:
> Yes, sorry I wasn't clearer but I meant to keep the `.gitpod.yml` file at the root level (because this is required, as you discovered) and then move the `.gitpod.Dockerfile` to `.gitpod/Dockerfile` (and so renaming it).

My latest commit is resolving the same

alecharp:
> My latest commit is resolving the same

yes I saw, you did that perfectly.

.gitpod/Dockerfile
alecharp: In this case, you need to remove the line

```
FROM gitpod/workspace-postgres
```

alecharp: Please read https://docs.docker.com/engine/reference/builder/#from and other resources about Docker and Dockerfile because here I'm not sure you see exactly what is happening with your instructions.

MondalSayantan: Thank you @alecharp for providing the documentation. After some research, I found that one should not put two base images in a single Dockerfile.  Docker will use only the `gitpod/workspace-full` base image and ignore the `gitpod/workspace-postgres` base image.

I have removed the `gitpod/workspace-full` base image as it is heavy and includes many tools and language support like `python` which might not be useful. 

`gitpod/workspace-postgres` contains support for Docker and has postgreSQL installed along with many utility tools for it. We are subsequently setting the environment for Java, Node.js and Maven. Yarn comes pre-installed with the image.

I have tested it by running on Gitpod.

alecharp:
@jmMeessen if you could share your wisdom on this subject, I would really appreciate. My question is: how is gitpod supposed to be used? Is this supposed to be a simple code editing environment or should it be capable to run the project?

.gitpod.yml
```diff
@@ -0,0 +1,7 @@
+image:
+  file: .gitpod/Dockerfile
+
+tasks:
+  - name: Build with Maven
+    command: mvn package -Dmaven.test.skip
+
```
alecharp, line 7:
```suggestion

```

.gitpod/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+RUN echo "nvm use default &>/dev/null" >> ~/.bashrc.d/51-nvm-fix
+
+# Install Maven
+RUN bash -c 'VERSION="3.9.0" &&\
+    . /home/gitpod/.sdkman/bin/sdkman-init.sh && \
+    sdk install maven $version && \
+    sdk default maven $version'
+    
+
```
alecharp, line 24:
```suggestion

```

alecharp (changes requested):

.gitpod/Dockerfile
MondalSayantan: Hi @alecharp, does my latest code change fix what you were hinting at?

jmMeessen:
Hello @MondalSayantan 

I need some more insight on the "why's" of the proposed setup. I'd like to know the reasons for requiring the postgres Gitpod base image. An explanation of the reasons of the various tooling choices would help. Note that the chosen Java JDK flavour is not the standard Jenkins uses (temurin).

If you want to have your own (private) customizations (ex: sidekicks like python or editor keystroke bindings) there are ways to achieve that. Don't forget that the choices made in the Gitpod configuration will impact all the potential users. 

Another important thing I wanted to make sure that you know (it could explain some of your choices): Gitpod allows to run Docker in Docker. Thus the docker-compose should work (didn't test it). No need to setup the "local" tooling for testing. Just leverage the power of Docker.

Just some thoughts. Interested to hear and understand the details of your proposal.

MondalSayantan:
Hello @jmMeessen 

> 
> I need some more insight on the "why's" of the proposed setup. I'd like to know the reasons for requiring the postgres Gitpod base image. An explanation of the reasons of the various tooling choices would help. Note that the chosen Java JDK flavour is not the standard Jenkins uses (temurin).
> 
On trying to open the project on Gitpod, the maven commands are failing to run because of incompatible Java version being used by default. I opened this PR to resolve all possible issues that may arise when using Gitpod to run and contribute to this project, to the best of my knowledge.

According to Gitpod documentation 'By default, Gitpod uses a standard Docker Image called `Workspace-full`. My choice of using `gitpod/workspace-postgres` is mostly because the image contains PostgreSQL database and some utility scripts along with it. I am replacing the default image, with another image which might be suitable to our needs.

I am using sdkman to change the Java version. I don't have any specific reason to use this tool.

I did not know about the Java JDK flavor. I will try to resolve it
> If you want to have your own (private) customizations (ex: sidekicks like python or editor keystroke bindings) there are ways to achieve that. Don't forget that the choices made in the Gitpod configuration will impact all the potential users.
> 
I am not customizing the workspace keeping my needs in mind. I am trying to provide the bare minimum tools and their correct versions to run the project.
> Another important thing I wanted to make sure that you know (it could explain some of your choices): Gitpod allows to run Docker in Docker. Thus the docker-compose should work (didn't test it). No need to setup the "local" tooling for testing. Just leverage the power of Docker.
> 
I knew about docker in docker, and I have tested it by running docker compose. The only local tool I am trying to add additionally is PostgreSQL (which might be required for development purposes). For the rest, I am only setting the correct versions. 
> Just some thoughts. Interested to hear and understand the details of your proposal.
>
Thank you for taking out the time to go through my PR. I hope I was able to convey my purpose behind it. Please correct me if you feel I am going in the wrong direction. I am learning as I am going :smiley:

jmMeessen:
Hello @MondalSayantan 

It starts to make sense 😄 Thank you for help me understand.

First of all, and I forgot to say it in my previous message 😐 , Gitpod is a very interesting technology. And it can help in many cases. So your PR could be a great addition to the project.

Based on my experience, I would avoid over-engineering the solution and keep the problems separate. In this case, I would first concentrate on having a workable  development and unit-test environment (== `mvn verify` working). 

As you know I am partial to using a docker environment for integration testing or interactive end-to-end testing.

What I suggest it to "rewind" your pull request to simplify it and tackle the first objective first.

I propose that you setup the minimalistic environment  (.gitpod that works). You can then have a set of private gitpod configuration scripts that will enhance/taylor it to your wishes/habits and allow to experiment with the full E2E (end-to-end) testing environment on gitpod.
Thus don't use the postgres base image and don't install node on the project's gitpod configuration.

I had the base environment work interactively with the base image. It requiers indeed to setup java 17, as you did. Just, instead of the microsoft JDK, use the Jenkins prefered JDK `sdk install java 17.0.6-tem` (interactive command).  

Yarn needs also to be installed. The default proposition by gitpod seems to work.

```
tasks:
  - init: yarn install && yarn run build && ./mvnw install -DskipTests=false
    command: yarn run dev
```

The default installed maven has the correct version

When this works, you could start another gitpod related PR where you concentrate on the E2E environment.

A nice addition (as relying blindly on defaults is not always a good idea) would be to create a small script (bash for linux and batch for windows) that checks that the pre-requiste (with the correct version) are available on the development environment.

Hoping that these thoughts and hints will be useful. Continue digging, @MondalSayantan and remember keep it simple...

/- Jmm

MondalSayantan:
Hello @jmMeessen 

I agree that I might have over-engineered the solution. I will simplify the changes, and tackle the most basic problem, i.e, setting the correct Java version and test the ability to run `mvn verify`, and then work further if required. 

Just a thought, `yarn run dev` command might not be required unless the contributor is looking to work on the frontend of the application. So I will add the `init` commands you proposed and leave out the commands after

Thank you so much for your insights!

alecharp:
With the latest `.gitpod.yml` file, it seems that the `Dockerfile` could be removed, correct?

alecharp:
> I did not add yarn run dev as it depends on the user's choice to run the project through CLI or Docker

Excellent choice.

> Should I squash these commits?

No don't worry, this repository is configured to only do squash merge from pull request exactly because pull request history can be messy and it is fine to figure out, fix things as they come up, improve and so on. 

Generally, it's also better to not touch the history of a pull request when some review were performed because it messes the reviews. This includes "forced pushed".

.gitpod.yml
```diff
@@ -0,0 +1,14 @@
...
+      yes Y | sdk install java 17.0.6-tem
+
+      # Download and Install Maven 3.9.0
+      yes Y | sdk install maven 3.9.0
+    init: |    
+      #Build using Maven
+      mvn package -Dmaven.test.skip
+      
+      # Install Yarn dependencies
+      yarn install
```
alecharp, line 14: Technically, this is done during the `generate-sources` phase of Maven (https://github.com/jenkins-infra/plugin-health-scoring/blob/main/pom.xml#L340-L349). I'm not sure this is strictly required as after the `mvn package`, the users would have downloaded the Yarn dependencies.

MondalSayantan: Thank you for letting me know

alecharp (approved):

jmMeessen:
That looks very elegant, @MondalSayantan .

I am running a little out of time right now and I want to give it adequate attention. I will look/test it tomorrow. 

I am also going to suggest an improvement in the configuration at the same time. It is a good idea that that the recommended VSC plugins are installed. (using Intellij with Gitpod requires the paid version). 

Once we have this step finalized, we can work on making the E2E test work in Gitpod and document this. It would be a nice addition for those with undersized dev configurations.

MondalSayantan:
>That looks very elegant, @MondalSayantan .
>
Thanks a lot!
> It is a good idea that that the recommended VSC plugins are installed.
>
I agree that it is a good idea to install some VSC extensions. 
The extensions VSC lists on their website as recommended for Java

- [Language Support for Java™ by Red Hat](https://marketplace.visualstudio.com/items?itemName=redhat.java)
- [Debugger for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-debug)
- [Test Runner for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-test)
- [Maven for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-maven)
- [Project Manager for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-dependency)
- [Visual Studio IntelliCode](https://marketplace.visualstudio.com/items?itemName=VisualStudioExptTeam.vscodeintellicode)

I will work on adding this suggestion.

jmMeessen (approved):
Nice work @MondalSayantan. And thank you for investing your time to add this (dev) feature.

Next step, if you are interested, is to make sure that the End to End test work in Gitpod and explain how to do it. I don't know this project but you got me interested in this 😉 .

alecharp:
we don't really have E2E testing but integration tests which use testcontainers to start a database container.

MondalSayantan:
> we don't really have E2E testing but integration tests which use testcontainers to start a database container.

Running `mvn verify` means running the test cases if I am not wrong. With the current `.gitpod.yml` configuration I am able to do that.

PR #211, opened 2023-02-13
Adds probe to detect existance of contributing guide on plugin repository

MondalSayantan:
This PR adds a probe to check the existence of a 'Contributing guide' in a plugin which may be of the form `CONTRIBUTING.md` or `CONTRIBUTING.adoc` anywhere in the repository.

**How is it helpful?**

- A contributing guide suggests that the plugin has an active and welcoming community that is open to new contributors.
- A contributing guide can outline the plugin's quality standards and expectations for contributions.

This probe provides valuable information for the **plugin health scoring project** by indicating whether the plugin has high-quality standards for code, testing, and documentation.

Probe suggested by @MarkEWaite


MondalSayantan:
Going forward, I will be working on the test cases and scoring for this probe

alecharp:
@MondalSayantan please note that the changelog of any release has links to pull requests which compose the release. Because of that, when anyone wants to understand the release, he/she can jump to the release. So it is very important that the pull request description is detailed. 

Please update your pull request description to explain why you are submitting this pull request, what problem you are trying to fix / improve, how it will contribute to the project. Thanks.

MondalSayantan:
> @MondalSayantan please note that the changelog of any release has links to pull requests which compose the release. Because of that, when anyone wants to understand the release, he/she can jump to the release. So it is very important that the pull request description is detailed.
> 
> Please update your pull request description to explain why you are submitting this pull request, what problem you are trying to fix / improve, how it will contribute to the project. Thanks.
>
Thank you @alecharp for pointing it out. I'll take care from the future

MondalSayantan:
I have created an unit test for the probe. This my first time exploring how unit tests work in Spring Boot, so I had to take reference from the `JenkinsfileProbeTest.java` which @alecharp authored.

MondalSayantan:
Hi @alecharp, I have added the probe and I think I can get started on scoring.

Could you provide some pointers on how I should proceed? Thanks

src/test/java/io/jenkins/pluginhealth/scoring/probes/ContributingGuidelinesProbeTest.java
```diff
@@ -0,0 +1,88 @@
...
+import java.nio.file.Paths;
+import java.time.ZonedDateTime;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+public class ContributingGuidelinesProbeTest {
```
alecharp, line 45: should the probe be executed when there are code changes on the repository? 
You need to test for that.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ContributingGuidelinesProbeTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void shouldCorrectlyDetectMissingContributingGuidelines() throws IOException {
+        final Plugin plugin = mock(Plugin.class);
+        final ProbeContext ctx = mock(ProbeContext.class);
+        final ContributingGuidelinesProbe probe = new ContributingGuidelinesProbe();
+
+        when(plugin.getDetails()).thenReturn(Map.of(
+                SCMLinkValidationProbe.KEY, new ProbeResult(SCMLinkValidationProbe.KEY, "", ResultStatus.SUCCESS,
+                        ZonedDateTime.now().minusMinutes(5))));
+        when(ctx.getScmRepository()).thenReturn(
+                Files.createTempDirectory("foo"));
```
alecharp, line 68: I'm pretty sure Checkstyle is pointing this as a problem.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ContributingGuidelinesProbeTest.java
```diff
@@ -0,0 +1,88 @@
...
+        final Plugin plugin = mock(Plugin.class);
+        final ProbeContext ctx = mock(ProbeContext.class);
+        final ContributingGuidelinesProbe probe = new ContributingGuidelinesProbe();
+
+        when(plugin.getDetails()).thenReturn(Map.of(
+                SCMLinkValidationProbe.KEY, new ProbeResult(SCMLinkValidationProbe.KEY, "", ResultStatus.SUCCESS,
+                        ZonedDateTime.now().minusMinutes(5))));
+        when(ctx.getScmRepository()).thenReturn(
+                Files.createFile(
+                        Paths.get(Files.createTempDirectory("foo").toAbsolutePath().toString(), "CONTRIBUTING.md")));
```
alecharp, line 84: Same as previous comment. Attend to the checkstyle report.

alecharp (changes requested):
Please see my comments

alecharp:
> Could you provide some pointers on how I should proceed? Thanks

I believe that this new probe should contribute to the `PluginMaintenanceScoring` class. 
As the already exists and has tests, please look at those. Overall, the scoring should:

 - be `0` is nothing is configured
 - having a Jenkinsfile should be 65% of the score
 - having dependabot and actually using it should be 15% of the score
 - having a contributing guide should be 15% of the score
 - having JEP-229 should be 5% of the score

Based on this, please change the tests (because this listing is changing how the scores will be affected) and then and only then change the code of the scoring implementation. @dheerajodha and I will review the tests first to make sure.

src/test/java/io/jenkins/pluginhealth/scoring/probes/ContributingGuidelinesProbeTest.java
MondalSayantan: I think I have a gap in my understanding. It would be really helpful if you could clarify it.

**What I think is happening**: Every two hours, the program is fetching all the plugins from the update center, and running all the probes on them which includes cloning of the repositories and searching of files like `CONTRIBUTING.md` and then modifying the scores automatically.

By 'code changes in the repository' I think you mean whenever someone merges a new code with the main branch of the plugin.

**My Question**: The present tests (inside the unit test file) can verify that the probe is able to find the contributing guidelines, why do we need to test if the probe is executing whenever there are code changes in the repository if the complete process is happening every two hours?

Also, `Jenkinsfile` probe unit test also does not contain anything similar due to which I am very confused.

alecharp: The problem is not the probe is or isn't able to find the file, but its conditions to be executed. 
In the `Probe` class, there are methods which help the `ProbeEngine` to determine if the probe should or shouldn't be executed:
 - `requiresRelease()`
 - `isSourceCodeRelated()`

It seems that I didn't add documentation on those methods so it's my fault. 

But, the first one specified that only a release could change the outcome of a probe on a plugin. So, if there is not new release since the last time a probe was executed on a plugin, there is no need to run the probe (the result would be the same as last time).

The second method specified that the probe requires code modifications to have a different outcome from the last execution.

By default, both methods are `false`. In your case, `isSourceCodeRelated` should be `true`, `requiresRelease` should be `false`.

I'll add documentation on those methods.

MondalSayantan: > But, the first one specified that only a release could change the outcome of a probe on a plugin. So, if there is not new release since the last time a probe was executed on a plugin, there is no need to run the probe (the result would be the same as last time).
> 
Just to be clear my confusion, if `requiresRelease` is `false`, it means the probe would run every time regardless if there is a new release or not.

alecharp: > Just to clear my confusion, if requiresRelease is false, it means the probe would run every time regardless if there is a new release or not as it is not limited by waiting for a new release.

correct.

> If my statement above is correct, then making isSourceCodeRelated true won't have any effect (is redundant) as the probe would be executed every time on code change based on my above statement.

So, ok, both condition could be redundant, as, when you have a release, you have code change that is true. But you can have source code change without release. That is why I created those two methods.

MondalSayantan:
In `PluginMaintenanceScoringTest` we have two separate methods `shouldScoreSeventyFiveForPluginsWithOnlyJenkinsfile()` and `shouldScoreSeventyFiveForPluginsWithJenkinsfileAndNoDependabot()` which are essentially doing the same thing.

Are they separate because in one case the `Dependabot` probe is not running and in the other case the probe is running but returning a failure?

alecharp:
> Are they separate because in one case the Dependabot probe is not running and in the other case the probe is running but returning a failure?

Yes because not having a probe result, having a failed probe result and having a successful probe result are three different possibilities which all need to be tested.

MondalSayantan:
These are all the tests
```
shouldScoreZeroWhenNoJenkinsfileProbeResult()
shouldScoreZeroForPluginsWithNoJenkinsfile()
shouldScoreSixtyFiveForPluginsWithOnlyJenkinsfile()
shouldScoreSixtyFiveForPluginsWithJenkinsfileAndNoDependabotAndNoContributingGuidelines()
shouldScoreSixtyFiveForPluginsWithJenkinsfileAndDependabotButOpenedPullRequest()
shouldScoreEightyForPluginsWithJenkinsfileAndContributingGuidelinesAndDependabotButOpenedPullRequests()l
shouldScoreEightyForPluginsWithOnlyJenkinsfileAndDependabotButNoContributingGuidelines()
shouldScoreEightyForPluginsWithJenkinsfileAndDependabotButNoJEP229AndNoContributingGuidelines()
shouldScoreNinetyFiveForPluginsWithJenkinsfileAndDependabotAndContributingGuidelines()
shouldScoreNinetyFiveForPluginsWithJenkinsfileAndDependabotAndContributingGuidelinesButNoJEP229()
shouldScoreHundredForPluginsWithJenkinsfileAndDependabotAndJEP229AndContributingGuidelines()
```
I was expecting shouldScoreHundredForPluginsWithJenkinsfileAndDependabotAndJEP229AndContributingGuidelines() to also fail, but this test case is passing, which I think is a loophole when I saw the implementation of scoring where there is no condition of `return new ScoreResult(KEY, 1, COEFFICIENT);`, where according to my opinion the conditions should be properly defined.

Please let me know if I am missing any test case.

MondalSayantan:
I removed the request for review by @dheerajodha by mistake.

src/test/java/io/jenkins/pluginhealth/scoring/scores/PluginMaintenanceScoringTest.java
```diff
@@ -46,135 +47,234 @@
...
-            DependabotPullRequestProbe.KEY, new ProbeResult(DependabotPullRequestProbe.KEY, "0", ResultStatus.SUCCESS),
-            ContinuousDeliveryProbe.KEY, new ProbeResult(ContinuousDeliveryProbe.KEY, "", ResultStatus.SUCCESS)
-        ));
-
-        final ScoreResult result = scoring.apply(plugin);
-        assertThat(result.key()).withFailMessage(() -> "Score key should be '%s'".formatted(KEY)).isEqualTo(KEY);
-        assertThat(result.coefficient()).withFailMessage(() -> "Score coefficient should be '%f'".formatted(COEFFICIENT)).isEqualTo(COEFFICIENT);
-        assertThat(result.value()).isEqualTo(1f);
-    }
+        public static final String KEY = "repository-configuration";
```
alecharp, line 50: The indentation of the entire code addition seems odd.

src/test/java/io/jenkins/pluginhealth/scoring/scores/PluginMaintenanceScoringTest.java
```diff
@@ -46,135 +47,234 @@
...
+        public static final float COEFFICIENT = .5f;
+
+        @Test
+        public void shouldScoreZeroWhenNoJenkinsfileProbeResult() {
```
alecharp, line 54: It's not only the Jenkinsfile that is missing in this case.
You could name that test as `shouldScoreZeroWithNoRepositoryToolsConfigured` or something like that.

alecharp (changes requested):
The indentation made the review a bit harder, but I think that all the cases were present. 
As all the tests are the same, modulo the probe results and the value, it could be great to use a parameterized test (see https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests with the `@MethodSource`), it could probably reduce the code, simplify the review and still cover everything we need.

MondalSayantan:
Working on this, I realized how important test cases are :)

src/main/java/io/jenkins/pluginhealth/scoring/scores/PluginMaintenanceScoring.java
```diff
@@ -46,25 +47,34 @@ protected ScoreResult doApply(Plugin plugin) {
...
         }
 
-        if (cdProbeResult == null || cdProbeResult.status().equals(ResultStatus.FAILURE)) {
-            return new ScoreResult(KEY, .9f, COEFFICIENT);
+        if (cdProbeResult != null && cdProbeResult.status().equals(ResultStatus.SUCCESS)) {
+            score += 0.05f;
         }
 
-        return new ScoreResult(KEY, 1, COEFFICIENT);
+        score = Math.round(score * 100.0) / 100.0f; // Prevent rounding errors
```
alecharp, line 75: Can you explain why you need this?

alecharp (changes requested):

src/main/java/io/jenkins/pluginhealth/scoring/scores/PluginMaintenanceScoring.java
MondalSayantan: Taking an example, `0.65f + 0.15f + 0.05f` was giving a result `0.84999996f` due to the way that these numbers are represented in binary. It was caught by the unit tests.

I am using `Math.round` method to round the result to two decimal places.

alecharp: Alright, this is clear enough for me. Thanks.

dheerajodha:
I'd like to review this tonight, can we please merge this tomorrow?

alecharp (approved):

PR #205, opened 2023-02-07
Detecting when a plugin is not published in the Update Center anymore

Jagrutiti:
# Closes

https://github.com/jenkins-infra/plugin-health-scoring/issues/178

# Changes 

All the changes are based on this comment: 

 https://github.com/jenkins-infra/plugin-health-scoring/issues/178#issuecomment-1413456490

> Here some pointers.
> 
> First, the condition of the probe:
> 
>     * it does not require a release to be updated
> 
>     * it does not require a code change to be updated
> 
> 
> Then, if you provide a `ProbeContext` with an `UpdateCenter` which doesn't contain the plugin in the map of `plugins`, then your probe should return a `FAILURE`, meaning the plugin is no longer published. If the plugin can be located in the map, then the probe should return a `SUCCESS` as the plugin is still published.
> 
> For the [`DeprecatedPluginScoring`](https://github.com/jenkins-infra/plugin-health-scoring/blob/main/src/main/java/io/jenkins/pluginhealth/scoring/scores/DeprecatedPluginScoring.java), for the moment, it's only checking the presence of the DeprecatedPluginProbe result. But it has now also to check the presence of the new probe.
> 
> I hope this clarify the requirements.

PR #212, opened 2023-02-14
Fixes update-center probe phrasing

alecharp:
This is coming from suggestions on #205 from @dheerajodha.


dheerajodha (approved):
Thank you for taking the time and opening this PR. Looks good!

PR #213, opened 2023-02-14
Reconfigure how new code issues are compared

alecharp:
### Description

I want to change how the code analytics are compared so that in a pull request, if we had a new problem in one commit, the next commit won't say there is nothing new. It needs to compare with the target branch of the pull request, not the last build of the pull request.

The selection of the baseline is configured using Git Forensics plugin, based on https://github.com/jenkinsci/warnings-ng-plugin/blob/master/doc/Documentation.md#selecting-a-baseline-in-the-target-job

### Submitter checklist

- [ ] If the issue exists, it is well described and linked in the description
    - `fix/` for bug fixes
- [ ] There is automated tests to cover the code change / addition
- [ ] There is no new warnings on the code


dheerajodha:
Just for my reference, if you don't mind, can you share the links that you referred to make these changes?

alecharp:
sorry that I didn't include those from the start.

alecharp:
You can see that the checks on https://github.com/jenkins-infra/plugin-health-scoring/pull/213/commits/a589862fd5e833c4e0b6309765d33359ac68c93c states that there are 4 checkstyle warnings but no new (when the commit created the warnings) and then https://github.com/jenkins-infra/plugin-health-scoring/pull/213/commits/00d7936f0bef96e58817e26bc9cf19370634e1a9 states that the warnings are new (because compared to `main`). 

This means the configuration works.

After that, I configured spotbugs to ignore issues created because of the usage of private local `record` and I fixed a real potential issue.

dheerajodha:
I'm just talking to myself here, I think I understood your point.

> there are 4 checkstyle warnings but no new (when the commit created the warnings)

So at this point, the "New" column only listed the warnings compared to the previous commit, and since this was the first commit hence the number "4" appears in the "Total" column.

> then https://github.com/jenkins-infra/plugin-health-scoring/commit/00d7936f0bef96e58817e26bc9cf19370634e1a9 states that the warnings are new (because compared to main)

Yes, that makes sense. Previously, there should be "0" under the "New" column because these 4 warnings have already been captured in the previous commit. But we see them here under the "New" column because we are comparing it from the main branch (as per the change you did in this commit).

dheerajodha (approved):

PR #221, opened 2023-02-24
chore: pin GitHub actions with their SHA-1 instead of their version number

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/3402


alecharp:
@lemeurherve I guess this one was a duplicate of #220?

alecharp:
As there is no diff from `main` now, I'm closing this.

PR #263, opened 2023-03-27
Fixes probe requirements to be executed

### LINKED ISSUE #262: Probes executed before their order
### Description of the bug

Probes about code coverage, jep-229, dependabot and jenkinsfile are executed when commits were added to the repository since the last execution. 
They are normally ordered to be executed after the last commit date is registered in the `ProbeContext` but apparently, this is not the case.

alecharp:
### Description

Resolves #262.

Since logging when a probe is supposed to be executed when a new code was added to the repository but the last commit date is not registered yet, we can see errors in the logs. 

Some plugin have an incorrect SCM link. The link is valid because the repository exists but they are pointed to a sub folder which doesn't exists. This causes `LastCommitDateProbe` to fail when `SCMLinkValidationProbe` is successful.

### Submitter checklist

  - `fix/` for bug fixes

### DIFF
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/CodeCoverageProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/CodeCoverageProbe.java
@@ -50,11 +50,6 @@ public class CodeCoverageProbe extends Probe {
 
     @Override
     protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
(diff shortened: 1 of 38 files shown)


alecharp:
Because of this discovery, I'm tempted to update the `PluginMaintenanceScoring` implementation to set a `0` when the `SCMLinkValidationProbe` is successful but not the `LastCommitDateProbe`.

PR #269, opened 2023-03-30
Changes ProbeResult#status meaning and centralize scoring

### LINKED ISSUE #264: Add scores and categories for individual probes
Help me understand what's a problem and how much, by providing, for individual probes:

* The top-level category they count towards
* Their individual score
* (If applicable) their weight of contribution to their top-level category

### LINKED ISSUE #268: `ProbeResult#status` is not reliable accross probes
### Description of the bug

Some probes should be deciding if the value they are testing is good enough to be considered the plugin as ok or not.

Also, when a probe is executed and an error occurs, we are keeping the previous result when in fact the error should discard the previous result.
This because, if a SCM link is now not valid (repository was deleted), then we don't want to keep the probe result as successful.

alecharp:
### Description

Resolves #268.
Resolves #264.

The probes are not expected to explicit their execution requirements. So, on top of the `@Order` annotation, a probe can be skipped if the probes required were not executed successfully on the plugin.

The scoring process can be simplified as they state the probe(s) required to evaluate a plugin, based only on the `ProbeResult#status` value of each probe.
This will also helped to describe why a plugin does or not get a full score for any scoring implementation.

I included the probe used for each scoring implementation and their status. This should help understanding the value of each scoring implementation. When the page of the plugin score is opened, the scoring implementation which don't have a full score (`1.0`) are by default expanded to show what is missing. Others are collapsed.

<details>
<summary>Before:</summary>

Screenshot 2023-03-31 at 17-00-01 Plugin Health Scoring Score
</details>

<details>
<summary>After:</summary>

https://user-images.githubusercontent.com/985955/229586117-f0aa76d1-e19f-4a74-a862-c2f37079501f.mov
</details>

### Submitter checklist

  - `fix/` for bug fixes


jleon33:
Very nice!

PR #271, opened 2023-04-02
Automatically remove old plugin scores

### LINKED ISSUE #193: Automatically removing old scores for each plugin
We can keep 5 or 10 scores for each plugins to get a trend, but we don't need to keep all the scores ever computed.
This needs to be scheduled and well tested.

- [ ] Removes scores for each plugins except the last 5 (by `computed_at`).
- [ ] Schedule this process to run every hour.
- [ ] This needs to be tested.

Sayo-gs:
resolved #193

### DIFF
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/service/ScoreService.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/service/ScoreService.java
@@ -68,6 +68,25 @@ public class ScoreService {
             ));
     }
 
+    @Transactional
+    public void deleteOldPluginScores() {
+        Map<String, List<Score>> scoresByPlugin = repository.findAll().stream.collect(Collectors.groupingBy(score -> score.getPlugin().getName(),Collectors.toList
+        ));
+
+        for(List<Score> scores = scoresByPlugin.values()) {
+            if(scores.size() > 5) {
+                List<Score> scoresToRemove = scores.stream.sorted((s1, s2) -> s2.getComputedAt().compareTo(s1.getComputedAt())).skip(5).collect(Collectors.toList());
+
+              repository.deleteAll(scoresToRemove);
+            }
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


Sayo-gs:
can you tell me why that 1 check is not passed

Sayo-gs:
Hey @alecharp 
can you guide me why that single check was not passed

core/src/main/java/io/jenkins/pluginhealth/scoring/service/ScoreService.java
```diff
@@ -68,6 +68,25 @@ public Map<String, ScoreSummary> getLatestScoresSummaryMap() {
...
+
+              repository.deleteAll(scoresToRemove);
+            }
+        }
+    }
+
+     @Scheduled(cron = "0 0 * * * *")
+    public void scheduledDeleteOldPluginScores() {
+        deleteOldPluginScores();
+    }
```
alecharp, line 88: I'd prefer this to be on a separate class, most probably in `war` module.

alecharp (changes requested):
The pull request template is here to guide you, as well as the contributing guide. It would have helped you. 

Please see https://github.com/jenkins-infra/plugin-health-scoring/pull/271/checks?check_run_id=12455363353

alecharp:
Because of https://github.com/jenkins-infra/plugin-health-scoring/pull/271/checks?check_run_id=12455363353

core/src/main/java/io/jenkins/pluginhealth/scoring/service/ScoreService.java
GauravFin: where should it be placed in between schedule or service folder in war module, most probably schedule folder

alecharp: it can be in the scoring scheduler yes.

PR #289, opened 2023-04-12
Update code coverage step

NotMyFault:
The `publishCoverage` step is deprecated in favor of `recordCoverage`. Take a look at the pipeline docs of more information: https://www.jenkins.io/doc/pipeline/steps/code-coverage-api/#publishcoverage-publish-coverage-report-deprecated.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,7 +42,7 @@ pipeline {
             allowEmptyResults: false,
             testResults: '**/target/surefire-reports/*.xml, **/target/failsafe-reports/*.xml'
           )
-          publishCoverage adapters: [jacocoAdapter(mergeToOneReport: true, path: '**/target/site/**/jacoco.xml')]
+          recordCoverage(tools: [[parser: 'JACOCO', pattern: '**/target/site/**/jacoco.xml', mergeToOneReport: true]], sourceCodeRetention: 'MODIFIED')
           recordIssues enabledForFailure: true,
             tools: [mavenConsole(), java(), javaDoc()]
           recordIssues enabledForFailure: true,


Jenkinsfile
```diff
@@ -42,7 +42,7 @@ pipeline {
             allowEmptyResults: false,
             testResults: '**/target/surefire-reports/*.xml, **/target/failsafe-reports/*.xml'
           )
-          publishCoverage adapters: [jacocoAdapter(mergeToOneReport: true, path: '**/target/site/**/jacoco.xml')]
+          recordCoverage(tools: [[parser: 'JACOCO', pattern: 'coverage/target/site/jacoco-aggregate/jacoco.xml', mergeToOneReport: true]], sourceCodeRetention: 'MODIFIED')
```
alecharp, line 45: I'm wondering about the pattern used? Shouldn't it be the same?
The build in https://ci.jenkins.io/job/Infra/job/plugin-health-scoring/job/PR-289/1/coverage/ seems to be using the version on the `main` branch.

alecharp (changes requested):
I'm not sure the pattern used to select the jacoco files should be changed. 
There is no `coverage` folder at the root level nor `jacoco-aggregate` in the modules.

Jenkinsfile
NotMyFault: Oh right, don't mind my bad copy/paste from updating core 😅

alecharp: No problem. Thanks for the contribution

alecharp (approved):
Thanks @NotMyFault

alecharp:
Looking at #287 checks, it seems that the "reference job" is not picked up with this change. I'll have a look soon.

PR #334, opened 2023-06-08
🚧 Scoring component composition

### LINKED ISSUE #333: Compose scoring components description
The current implementation of a scoring components description limits the composition of probe results.
We will need in the future to be able to express composition like a boolean expression to say "To get this many points, the plugin must have both this probe result valid and this other one".

The current `PluginMaintenanceScoring` implementation is doing so by having negative points to the secondary requirement, matching the points of the first one. See https://github.com/jenkins-infra/plugin-health-scoring/blob/d2d54972f01badc653cc33016b8b90d3574411a1/core/src/main/java/io/jenkins/pluginhealth/scoring/scores/PluginMaintenanceScoring.java#L57-L58. 

However, this won't be sufficient when we have to work on `RenovateProbe` (in #254) for example.

# Requirement

Current, we need to be able to express:
 - a direct requirement (probe's key)
 - a `and` expression between two direct requirement

alecharp:
### Description

Resolves #333.

This creates an architecture where we can express boolean expression to score a plugin based on the presence of successful probe results in a plugin detail.

```[tasklist]
### Submitter checklist
- [ ] Run `mvn verify` locally and all tests are passing successfully
- [ ] There is no new warnings (checkstyle nor spotbugs) on the code
```

### DIFF
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/scores/AdoptionScoring.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/scores/AdoptionScoring.java
@@ -24,16 +24,8 @@
 
 package io.jenkins.pluginhealth.scoring.scores;
 
-import java.time.Duration;
-import java.time.ZonedDateTime;
-import java.time.temporal.ChronoUnit;
 import java.util.Map;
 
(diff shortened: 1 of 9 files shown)


alecharp:
I'm taking a more drastic refactoring approach, thanks to @aneveux

PR #342, opened 2023-06-18
Security scan workflow probe

### LINKED ISSUE #257: Add a probe for the Jenkins security scan workflow
### Description

It would be nice to have a probe for the Jenkins security scan, https://github.com/search?q=org%3Ajenkinsci+%22uses%3A+jenkins-infra%2Fjenkins-security-scan%22&type=code

I included the workflow in the template new plugin hosting requests use, therefore the overall usage will increase within new plugins 🎉

Jagrutiti:
### Description

This probe checks whether the workflow defined in `jenkins-infra/jenkins-security-scan/.github/workflows/jenkins-security-scan.yaml` is configured in the plugin repository.

Closes #257 

### Testing done

- Unit test cases in progress.

### Submitter checklist


core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanInGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(value = SCMLinkValidationProbe.ORDER)
+public class SecurityScanInGitHubWorkflowProbe extends Probe {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SecurityScanInGitHubWorkflowProbe.class);
+    public static final int ORDER = SCMLinkValidationProbe.ORDER + 100;
+    public static final String KEY = "security-scan-probe";
```
alecharp, line 41: non of the probe have a key with `-probe` suffix.

```suggestion
    public static final String KEY = "security-scan";
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanInGitHubWorkflowProbeTest.java
```diff
@@ -0,0 +1,61 @@
...
+        final SecurityScanInGitHubWorkflowProbe probe = getSpy();
+
+        when(plugin.getDetails()).thenReturn(Map.of(
+            SCMLinkValidationProbe.KEY, ProbeResult.success(SCMLinkValidationProbe.KEY, "")
+        ));
+        final Path repo = Files.createTempDirectory("foo");
+
+        when(ctx.getScmRepository()).thenReturn(repo);
+        Files.createDirectories(repo.resolve(".github/workflows"));
+        when(ctx.getScmRepository()).thenReturn(repo);
```
alecharp, line 53:
```suggestion
```

alecharp (changes requested):

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanInGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,62 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(value = SCMLinkValidationProbe.ORDER)
+public class SecurityScanInGitHubWorkflowProbe extends Probe {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SecurityScanInGitHubWorkflowProbe.class);
```
alecharp, line 39:
```suggestion
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanInGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,62 @@
...
+ * SOFTWARE.
+ */
+
+package io.jenkins.pluginhealth.scoring.probes;
+
+import io.jenkins.pluginhealth.scoring.model.Plugin;
+import io.jenkins.pluginhealth.scoring.model.ProbeResult;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
alecharp, line 31:
```suggestion
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Test
+    void shouldBeAbleToDetectRepositoryWithNoGitHubWorkflowConfigured() throws IOException {
+        final Plugin plugin = mock(Plugin.class);
+        final ProbeContext ctx = mock(ProbeContext.class);
+        final SecurityScanGithubWorkflowProbe probe = getSpy();
+
+        when(plugin.getDetails()).thenReturn(Map.of(
+            SCMLinkValidationProbe.KEY, ProbeResult.success(SCMLinkValidationProbe.KEY, "")
+        ));
```
alecharp, line 56: I don't think this is enough based on https://github.com/jenkins-infra/plugin-health-scoring/pull/342/files#diff-f623053886a3b527e504f391ddc438541068ae28e1ad69db06f058a05f0ad33fR94

alecharp (changes requested):
Please see if my comment is applicable elsewhere

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,86 @@
...
+import java.nio.file.Path;
+import java.util.Map;
+
+import io.jenkins.pluginhealth.scoring.model.Plugin;
+import io.jenkins.pluginhealth.scoring.model.ProbeResult;
+import io.jenkins.pluginhealth.scoring.model.ResultStatus;
+
+import org.junit.jupiter.api.Test;
+
+public class SecurityScanGithubWorkflowProbeTest extends AbstractProbeTest<SecurityScanGithubWorkflowProbe> {
```
alecharp, line 43: Test classes should be package visible only.

```suggestion
class SecurityScanGithubWorkflowProbeTest extends AbstractProbeTest<SecurityScanGithubWorkflowProbe> {
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+        Files.write(workflowFile, List.of(SecurityScanGithubWorkflowProbe.SEARCH_LINE));
+        when(ctx.getScmRepository()).thenReturn(repo);
+
+        final ProbeResult result = probe.apply(plugin, ctx);
+        assertThat(result.status()).isEqualTo(ResultStatus.FAILURE);
+        assertThat(result.message()).isEqualTo("GitHub workflow security scan is not configured in the plugin");
+    }
+
+    @Test
+    void shouldFailIfUsesIsNotConfigured() throws IOException {
```
Jagrutiti, line 131: Should this test succeed of fail?

PierreBtz: A test you commit should always succeed (except if you are committing a failing test on purpose to prove a bug, in which case you would deactivate the test so that it doesn't break the build, with something like `@Disabled("Reproduces my bug")`).

What `fail` means in this test method is that the probe should fail, as evidenced by the assertions:

```
assertThat(result.status()).isEqualTo(ResultStatus.FAILURE);
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+        final ProbeContext ctx = mock(ProbeContext.class);
+        final SecurityScanGithubWorkflowProbe probe = getSpy();
+
+        when(plugin.getDetails()).thenReturn(Map.of(
+            SCMLinkValidationProbe.KEY, ProbeResult.success(SCMLinkValidationProbe.KEY, ""),
+            LastCommitDateProbe.KEY, ProbeResult.success(LastCommitDateProbe.KEY, "")
+        ));
+        final Path repo = Files.createTempDirectory("foo");
+        Path workflowPath = Files.createDirectories(repo.resolve(".github/workflows"));
+        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml@v2"));
```
alecharp, line 120: Not a "valid" file. Extension of the file `.yaml@v2` is not something that exists. It has to be `.yaml` or `.yml`.

```suggestion
        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml"));
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+        Files.write(workflowFile, List.of(SecurityScanGithubWorkflowProbe.SEARCH_LINE));
```
alecharp, line 122: This won't create a valid workflow definition file. You are creating an invalid workflow definition syntax.

Please see https://github.com/jenkins-infra/plugin-health-scoring/blob/d4b09077a456df2a1f2363fa21fccd722042ccc6/core/src/test/java/io/jenkins/pluginhealth/scoring/probes/ContinuousDeliveryProbeTest.java#L110-L115

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+        when(ctx.getScmRepository()).thenReturn(repo);
+
+        final ProbeResult result = probe.apply(plugin, ctx);
+        assertThat(result.status()).isEqualTo(ResultStatus.FAILURE);
+        assertThat(result.message()).isEqualTo("GitHub workflow security scan is not configured in the plugin");
+    }
+
+    @Test
+    void shouldFailIfJobsIsNotConfiguredInGitHubWorkflow() throws IOException {
```
alecharp, line 109: In this test, you create a file in the `.github/workflows` and put in this files (incorrectly but still) the value you are looking for. 
I would expect that this test to be looking for a `ResultStatus.SUCCESS`.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+            LastCommitDateProbe.KEY, ProbeResult.success(LastCommitDateProbe.KEY, "")
+        ));
+        final Path repo = Files.createTempDirectory("foo");
+        Path workflowPath = Files.createDirectories(repo.resolve(".github/workflows"));
+        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml@v2"));
+
+        Files.write(workflowFile, List.of(
+            "jobs:",
+            "  security-scan:",
+            "    this-is-not-uses: jenkins-infra/jenkins-security-scan/.github/workflows/jenkins-security-scan.yaml@v2"
```
alecharp, line 147: I wouldn't do that as you are creating a file with an incorrect syntax.

```suggestion
            "    uses: not-the-workflow-definition-we-are-lookgin-for@v1"
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+        final ProbeContext ctx = mock(ProbeContext.class);
+        final SecurityScanGithubWorkflowProbe probe = getSpy();
+
+        when(plugin.getDetails()).thenReturn(Map.of(
+            SCMLinkValidationProbe.KEY, ProbeResult.success(SCMLinkValidationProbe.KEY, ""),
+            LastCommitDateProbe.KEY, ProbeResult.success(LastCommitDateProbe.KEY, "")
+        ));
+        final Path repo = Files.createTempDirectory("foo");
+        Path workflowPath = Files.createDirectories(repo.resolve(".github/workflows"));
+        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml@v2"));
```
alecharp, line 142: dito
```suggestion
        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml"));
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
```
alecharp, line 168: same

```suggestion
        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml"));
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+        ));
+        when(ctx.getScmRepository()).thenReturn(repo);
+
+        final ProbeResult result = probe.apply(plugin, ctx);
+        assertThat(result.status()).isEqualTo(ResultStatus.FAILURE);
+        assertThat(result.message()).isEqualTo("GitHub workflow security scan is not configured in the plugin");
+    }
+
+    @Test
+    void shouldFailIfSecurityScanIsNotConfigured() throws IOException {
```
alecharp, line 157: This test should expect a `ResultStatus.SUCCESS`.
You are putting the valid workflow definition in the file. See https://github.com/jenkins-infra/plugin-health-scoring/pull/342/files#diff-ec875a98fd9e4055806edf694288fad916442221364b894d56c3c3917450bc0fR173.

We should see that the security scan workflow is configured.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+        final ProbeContext ctx = mock(ProbeContext.class);
+        final SecurityScanGithubWorkflowProbe probe = getSpy();
+
+        when(plugin.getDetails()).thenReturn(Map.of(
+            SCMLinkValidationProbe.KEY, ProbeResult.success(SCMLinkValidationProbe.KEY, ""),
+            LastCommitDateProbe.KEY, ProbeResult.success(LastCommitDateProbe.KEY, "")
+        ));
+        final Path repo = Files.createTempDirectory("foo");
+        Path workflowPath = Files.createDirectories(repo.resolve(".github/workflows"));
+        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml@v2"));
```
alecharp, line 194: dito

```suggestion
        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml"));
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+        Files.write(workflowFile, List.of(
+            "jobs:",
+            "  this-is-not-security-scan:",
```
alecharp, line 172: This is just a "name", it has not effect on the fact that the workflow declared in the next line is or is not the security scan workflow that we are looking for.

This is why the test should be expected in a `ResultStatus.SUCCESS`.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+        final ProbeContext ctx = mock(ProbeContext.class);
+        final SecurityScanGithubWorkflowProbe probe = getSpy();
+
+        when(plugin.getDetails()).thenReturn(Map.of(
+            SCMLinkValidationProbe.KEY, ProbeResult.success(SCMLinkValidationProbe.KEY, ""),
+            LastCommitDateProbe.KEY, ProbeResult.success(LastCommitDateProbe.KEY, "")
+        ));
+        final Path repo = Files.createTempDirectory("foo");
+        Path workflowPath = Files.createDirectories(repo.resolve(".github/workflows"));
+        final Path workflowFile = Files.createFile(workflowPath.resolve("incorrect-file-name.yml"));
```
alecharp, line 98: This is not an incorrect filename.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+        Files.write(workflowFile, List.of(SecurityScanGithubWorkflowProbe.SEARCH_LINE));
+        when(ctx.getScmRepository()).thenReturn(repo);
+
+        final ProbeResult result = probe.apply(plugin, ctx);
+        assertThat(result.status()).isEqualTo(ResultStatus.FAILURE);
+        assertThat(result.message()).isEqualTo("GitHub workflow security scan is not configured in the plugin");
+    }
```
alecharp, line 106:
```suggestion
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
alecharp: Well, to show a bug, I would said that "breaking the build" is what we could expect from fixing a test or adding one. I, personally, have never been a great supporter of `@Disabled` because this could be forgotten.

In the current situation, you (@Jagrutiti) are simply creating an invalid workflow definition syntax so yes, the test should be looking for a `ResultStatus.FAILURE`. 
The test shouldn't be named `shouldFail` but `shouldNotFindWhatWeAreLookingFor` (example, don't use that).

alecharp (changes requested):
Sorry, last review was definitively not supposed to be an approval

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
Jagrutiti: You mean `incorrect-file-name.yml` is a correct file name?

alecharp: Yes. The name of the file has nothing to do with its content. Of course, we tend to name them accordingly. But, you could name a file `security-scan.yml` and configure an hello world in it.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
Jagrutiti: This should succeed because the file is present?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,196 @@
...
+        final Path repo = Files.createTempDirectory("foo");
+        Path workflowPath = Files.createDirectories(repo.resolve(".github/workflows"));
+        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml"));
+
+        Files.write(workflowFile, List.of(
+            "name: Test Security Scan Job",
+            "nojobs-configured:",
+            "  security-scan-name:",
+            "    uses: jenkins-infra/jenkins-security-scan/.github/workflows/jenkins-security-scan.yaml"
+        ));
```
alecharp, line 105: Same as before. This is creating an invalid file syntax. This is not the way you should do it.
To be honest, when I read the test, I thought this test was vastly incorrect as you are creating a YAML file in the correct folder with the workflow we are looking for, so we should expect a `ResultStatus.SUCCESS`. 

```suggestion
        Files.write(workflowFile, List.of(
            "name: Test Security Scan Job",
            "jobs:",
            "  security-scan-name:",
            "    uses: this-is-not-the-workflow-we-are-looking-for"
        ));
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,196 @@
...
+        Files.createDirectories(repo.resolve(".github/workflows"));
+        when(ctx.getScmRepository()).thenReturn(repo);
+
+        final ProbeResult result = probe.apply(plugin, ctx);
+        assertThat(result.status()).isEqualTo(ResultStatus.FAILURE);
+        assertThat(result.message()).isEqualTo("GitHub workflow security scan is not configured in the plugin");
+    }
+
+    @Test
+    void shouldFailIfJobsIsNotConfiguredInGitHubWorkflow() throws IOException {
```
alecharp, line 87:
```suggestion
    void shouldNotFindSecurityScanConfiguredInGitHubWorkflow() throws IOException {
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,196 @@
...
+            "jobs:",
+            "  security-scan:",
+            "    uses: not-the-workflow-definition-we-are-looking-for@v1"
+        ));
+        when(ctx.getScmRepository()).thenReturn(repo);
+
+        final ProbeResult result = probe.apply(plugin, ctx);
+        assertThat(result.status()).isEqualTo(ResultStatus.FAILURE);
+        assertThat(result.message()).isEqualTo("GitHub workflow security scan is not configured in the plugin");
+    }
```
alecharp, line 138:
```suggestion
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,196 @@
...
+        ));
+        final Path repo = Files.createTempDirectory("foo");
+        Path workflowPath = Files.createDirectories(repo.resolve(".github/workflows"));
+        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml"));
+
+        Files.write(workflowFile, List.of(
+            "name: Test Security Scan Job",
+            "jobs:",
+            "  security-scan:",
+            "    uses: jenkins-infra/jenkins-security-scan/.github/workflows/jenkins-security-scan.yaml"
```
alecharp, line 185: Just to make sure that the version is not a problem for the code.

```suggestion
            "    uses: jenkins-infra/jenkins-security-scan/.github/workflows/jenkins-security-scan.yaml@v42"
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,196 @@
...
+        ));
+        when(ctx.getScmRepository()).thenReturn(repo);
+
+        final ProbeResult result = probe.apply(plugin, ctx);
+        assertThat(result.status()).isEqualTo(ResultStatus.SUCCESS);
+        assertThat(result.message()).isEqualTo("GitHub workflow security scan is configured in the plugin");
+    }
+
+    @Test
+    void shouldSucceedIfWorkflowIsConfigured() throws IOException {
```
alecharp, line 168:
```suggestion
    void shouldSucceedToFindWorkflowEvenWithVersion() throws IOException {
```

alecharp (changes requested):

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+
+                return files
+                    .map(file -> {
+                        try {
+                            return yaml.readValue(Files.newInputStream(file), SecurityScanGithubWorkflowProbe.WorkflowDefinition.class);
+                        } catch (IOException e) {
+                            LOGGER.error("Couldn't not read {} for {} on {}", file, key(), plugin.getName(), e);
+                            return new SecurityScanGithubWorkflowProbe.WorkflowDefinition(Map.of());
+                        }
+                    })
```
aneveux, line 77: For better readability, I would advise that you extract this piece of code in a separate function. 

This way you write proper Javadoc for that function, possibly unit test it if needs be.

Then your code also is easier to read cause you could have your first map reading like:

`files.map(::parseYaml)` or `readYaml` for example.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+                    .filter(wf -> wf.jobs() != null && !wf.jobs().isEmpty())
+                    .flatMap(wf -> wf.jobs().values().stream())
```
aneveux, line 79: While it would seem excessive to extract this in a separate function (debatable, but it's just my feeling), a small one line comment simply describing what exactly is being retrieved here could be helpful.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+                    .map(SecurityScanGithubWorkflowProbe.WorkflowJobDefinition::uses)
+                    .filter(Objects::nonNull)
+                    .anyMatch(def -> def.startsWith(SEARCH_LINE)) ?
```
aneveux, line 82: Is there a particular reason to use `startsWith` and not `equals` or `contains` for example? 

Also again, what is `def`?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+                    ProbeResult.success(key(), "GitHub workflow security scan is configured in the plugin") :
+                    ProbeResult.failure(key(), "GitHub workflow security scan is not configured in the plugin");
```
aneveux, line 84: The ternary operator does not bring a lot in that piece of code, and IMHO actually makes the source code less trivial to read. Each line before is an operation on a `Stream`, but the 2 last lines with similar indentation are actually the return result of the function.

I would advise to store the result of the Stream operations in a properly named variable, and then use it in a return statement.

If you want to use a ternary operator, you could use it with `return workflowsContainSecurityScanner ? success : failure` for example.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+
+        } catch (IOException e) {
+            return ProbeResult.error(key(), e.getMessage());
```
aneveux, line 87: No log message for this one?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    private record WorkflowJobDefinition(String uses) {
```
aneveux, line 92: This deserves Javadoc, to explain what that `record` is for, and ideally with a link to the syntax reference of GitHub workflows, so future developers understand where `uses` comes from.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    private record WorkflowDefinition(Map<String, SecurityScanGithubWorkflowProbe.WorkflowJobDefinition> jobs) {
```
aneveux, line 96: Similar comment for the Javadoc need.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,169 @@
...
+import org.junit.jupiter.api.Test;
+
+class SecurityScanGithubWorkflowProbeTest extends AbstractProbeTest<SecurityScanGithubWorkflowProbe> {
+    @Override
+    SecurityScanGithubWorkflowProbe getSpy() {
+        return spy(SecurityScanGithubWorkflowProbe.class);
+    }
+
+    @Test
+    void shouldBeAbleToDetectRepositoryWithNoGitHubWorkflowConfigured() throws IOException {
```
aneveux, line 51: FTR, if you want to have human readable test names, you can use `@DisplayName("Human Readable Name")`.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
```diff
@@ -0,0 +1,169 @@
...
+        ));
+        final Path repo = Files.createTempDirectory("foo");
+        Path workflowPath = Files.createDirectories(repo.resolve(".github/workflows"));
+        final Path workflowFile = Files.createFile(workflowPath.resolve("jenkins-security-scan.yaml"));
+
+        Files.write(workflowFile, List.of(
+            "name: Test Security Scan Job",
+            "jobs:",
+            "  security-scan-name:",
+            "    uses: this-is-not-the-workflow-we-are-looking-for"
```
aneveux, line 104: FTR, you can now use the `"""` operator in Java for multilines Strings, if you want something more readable where you don't have to repeat boilerplate characters.

aneveux (commented):
Hello :wave: 

Good job on the PR so far!

I left some comments, mostly about code readability, and some questions. Obviously feel free to challenge anything I say here, I don't have all the context you have in mind so it might be missing information.

Overall comment about the tests: you did not test a situation where there are multiple files in the workflows folder.

There is also some duplicated code in the tests. Feel free to extract similar pieces in helper functions that you call while setting up your tests. 

Let me know if you have any questions about my remarks.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(SecurityScanGithubWorkflowProbe.ORDER)
+public class SecurityScanGithubWorkflowProbe extends Probe {
+    public static final int ORDER = LastCommitDateProbe.ORDER + 100;
```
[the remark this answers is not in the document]
alecharp: This is coming from Spring Framework. See https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/annotation/Order.html

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+    public static final String KEY = "security-scan";
+    public static final String SEARCH_LINE = "jenkins-infra/jenkins-security-scan/.github/workflows/jenkins-security-scan.yaml";
```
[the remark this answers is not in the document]
alecharp: To be fair, I don't really like it to be public and I'd like to see this class be redesigned to have a generic one with the `ContinuousDeliveryProbe`.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final String WORKFLOWS_DIRECTORY = ".github/workflows";
+    private static final Logger LOGGER = LoggerFactory.getLogger(SecurityScanGithubWorkflowProbe.class);
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        final Path repository = context.getScmRepository();
+        final Path workflowPath = Paths.get(repository.toString(), WORKFLOWS_DIRECTORY);
```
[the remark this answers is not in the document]
alecharp: Yes this would be better.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
alecharp: It's a "convention" I took from Javascript testing framework, to name test method `shouldXXX`.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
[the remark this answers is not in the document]
aneveux: Yeah, for the annotation I got it, I was rather wondering about how we are defining the values in there? Is there some kind of meaning/strategy in the way you define the `ORDER` values in variables?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbeTest.java
alecharp: with `Files.writeString` method then. Otherwize, `write` takes an iterable.

aneveux: Yep sure.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
alecharp: Not to worry about those right now as they should be de-duplicated first.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,119 @@
...
+
+        if (! Files.exists(workflowPath)) {
+            return ProbeResult.failure(key(), "GitHub workflow directory could not be found in the plugin");
+        }
+
+            try (Stream<Path> files = Files.find(workflowPath, 1,
+                (path, basicFileAttributes) -> Files.isRegularFile(path)
```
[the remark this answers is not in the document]
Jagrutiti: Special character as a variable name? I always had in mind that special character cannot be used as variable names. 

> If that approach doesn't suit you, then I would suggest naming it unused to make it clear it's a parameter that you don't care about.

I will implement the way you showed me. It is interesting.

core/src/main/java/io/jenkins/pluginhealth/scoring/utility/GitHubWorkflowReader.java
alecharp, line 1: This is not quite what I had in mind. 
The idea is to create a class, which would inherit from `Probe` and which would take most of the `doApply` method code from the `ContinousDeliveryProbe` and `SecurityScanGithubWorkflowProbe` classes. It would defined an abstract method (let's say `getWorkflowDefinition`) which returns a `String` and that method can be used in place the of the current `anyMatch` line of the stream code.

In the end, `ContinuousDeliveryProbe` and `SecurityScanGithubWorkflowProbe` would extend that class, and only have to implement `getWorkflowDefinition` method and probably 2 others (to get the message if the workflow is found or not).

alecharp: Don't forget to add the license header.

Jagrutiti: Do you want me to replace only the following line?

` ...anyMatch(def -> def.startsWith("jenkins-infra/github-reusable-workflows/.github/workflows/maven-cd.yml")) ...`

With the abstract method? 

The rest of the code in the `doApply()` of `ContinousDeliveryProbe` and `SecurityScanGithubWorkflowProbe ` remains the same?

Jagrutiti: I find these two statements contradicting each other: 

> which would take most of the `doApply` method code from the `ContinousDeliveryProbe` and `SecurityScanGithubWorkflowProbe` classes.


and


> which returns a `String` and that method can be used in place the of the current `anyMatch` line of the stream code.

That is why I am confused here.

Jagrutiti: Hi

I have added a new class to remove redundant code and implemented it in `ContinuousDeliveryProbe`.

If my approach is correct, I will extend it to `SecurityScanGithubWorkflowProbe`.

Let me know your thoughts.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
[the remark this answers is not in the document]
alecharp: ok so no it's not documented. 
The idea was rather than having increment of 1, I choose to go with increment of 100 so that when listing the probes, the order is more obvious.
The increment of 10 are reserved to probes that have a strict link with its predecessor.

The idea is also to say "probe B needs probe A, so increment probe A by 100 for probe B". If probe C also needs probe A, do the same. If probe C also need probe B then let's use probe B as the base.

This may make only sense in my head..

core/src/main/java/io/jenkins/pluginhealth/scoring/utility/GitHubWorkflowReader.java
alecharp, line 1: This class shouldn't be in `utility` package but in `probes` package.
You should also rename it `AbstractGitHubWorkflowProbe` as this is not a reader.

core/src/main/java/io/jenkins/pluginhealth/scoring/utility/GitHubWorkflowReader.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package io.jenkins.pluginhealth.scoring.utility;
+
+import java.nio.file.Path;
+
+import io.jenkins.pluginhealth.scoring.probes.Probe;
+
+public abstract class GitHubWorkflowReader extends Probe {
+    public abstract String getWorkflowDefinition(Path githubWorkflow);
```
alecharp, line 32: This shouldn't take a parameter and have a javadoc.
```suggestion
    /**
     * Returns the path to the GitHub Workflow definition which should be use in one of the actions of the plugin repository.
     *
     * @return the workflow definition used in one of the jobs of one of the actions of the plugin repository.
     */
    public abstract String getWorkflowDefinition();
```

core/src/main/java/io/jenkins/pluginhealth/scoring/utility/GitHubWorkflowReader.java
```diff
@@ -0,0 +1,34 @@
...
+ * SOFTWARE.
+ */
+
+package io.jenkins.pluginhealth.scoring.utility;
+
+import java.nio.file.Path;
+
+import io.jenkins.pluginhealth.scoring.probes.Probe;
+
+public abstract class GitHubWorkflowReader extends Probe {
```
alecharp, line 31:
```suggestion
public abstract class AbstractGitHubWorkflowProbe extends Probe {
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,117 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(SecurityScanGithubWorkflowProbe.ORDER)
+public class SecurityScanGithubWorkflowProbe extends Probe {
```
alecharp, line 47:
```suggestion
public class SecurityScanGithubWorkflowProbe extends AbstractGitHubWorkflowProbe {
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ContinuousDeliveryProbe.java
```diff
@@ -28,11 +28,12 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Map;
-import java.util.Objects;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import io.jenkins.pluginhealth.scoring.model.Plugin;
 import io.jenkins.pluginhealth.scoring.model.ProbeResult;
+import io.jenkins.pluginhealth.scoring.utility.GitHubWorkflowReader;
```
alecharp, line 36:
```suggestion
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ContinuousDeliveryProbe.java
```diff
@@ -44,19 +45,29 @@
 
 @Component
 @Order(ContinuousDeliveryProbe.ORDER)
-public class ContinuousDeliveryProbe extends Probe {
-    private static final Logger LOGGER = LoggerFactory.getLogger(ContinuousDeliveryProbe.class);
+public class ContinuousDeliveryProbe extends GitHubWorkflowReader {
```
alecharp, line 48:
```suggestion
public class ContinuousDeliveryProbe extends AbstractGitHubWorkflowProbe {
```

alecharp (changes requested):
Please see that both `ContinuousDeliveryProbe` and `SecurityScanGithubWorkflowProbe` share a lot in common and all that should be removed from those classes and placed in the new abstract class.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Returns the path to the GitHub Workflow definition which should be use in one of the actions of the plugin repository.
+     *
+     * @return the workflow definition used in one of the jobs of one of the actions of the plugin repository.
+     */
+    public abstract String getWorkflowDefinition();
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    record WorkflowDefinition(Map<String, WorkflowJobDefinition> jobs) {
```
alecharp, line 103:
```suggestion
    private record WorkflowDefinition(Map<String, WorkflowJobDefinition> jobs) {
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    record WorkflowJobDefinition(String uses) {
```
alecharp, line 107:
```suggestion
    private record WorkflowJobDefinition(String uses) {
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public abstract String getFailureMessage();
+
+    public abstract String getSuccessMessage();
```
alecharp, line 112: Would be great to add the javadoc on those methods to know what they are used for and what they should return.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+import io.jenkins.pluginhealth.scoring.model.Plugin;
+import io.jenkins.pluginhealth.scoring.model.ProbeResult;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class AbstractGitHubWorkflowProbe extends Probe {
```
alecharp, line 43: Would be great to have javadoc on this to understand what is the purpose of this class.

alecharp (changes requested):

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+
+            return files
+                .map(file -> {
+                    try {
+                        return yaml.readValue(Files.newInputStream(file), WorkflowDefinition.class);
+                    } catch (IOException e) {
+                        LOGGER.warn("Couldn't not read {} for {} on {}", file, key(), plugin.getName(), e);
+                        return new WorkflowDefinition(Map.of());
+                    }
+                })
```
aneveux, line 71: I would suggest to extract this code block to a dedicated function, so you can Javadoc that function and explain properly what this bit does, and you can then use it in a `map` as a one liner to ease the reading.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+                .filter(wf -> wf.jobs() != null && !wf.jobs().isEmpty())
```
aneveux, line 72: Code is more often read than written, so it is worth the extra few characters to write proper entire names for things.

`workflow` is easier to read than `wf`, and does not make the code line too long, so I would suggest to use that.

Also, while it may feel over the top to extract this bit in a dedicated function to clearly indicate what you're doing, a one-line comment could be worth it to explain why this filtering?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+                .flatMap(wf -> wf.jobs().values().stream())
```
aneveux, line 73: Similar comment for the naming here. `workflow` reads better than `wf`.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+                .map(WorkflowJobDefinition::uses)
+                .filter(Objects::nonNull)
+                .anyMatch(def -> def.startsWith(getWorkflowDefinition())) ?
```
aneveux, line 76: Again for naming, I would suggest to rename `def` to something more accurate and self-descriptive. I believe here you are working only on all the information contained in the `uses` field of workflows, so perhaps something like `usedWorkflow` or something in that idea?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+                ProbeResult.success(key(), getSuccessMessage()) :
+                ProbeResult.failure(key(), getFailureMessage());
```
aneveux, line 78: I feel like the ternary operator is not helping with readability here, because it indents at the same level as all the operations you are doing on your stream.

I believe that extracting the result of the Streams operations to a variable, something like `specificWorkflowUsageFound` for example, would help readability.

You can then write a log message with that information, and use a ternary operation just for returning your probe result, with something like:

`return specificWorkflowUsageFound ? success : failure`

Again, for readability purposes.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+
+        } catch (IOException e) {
+            return ProbeResult.error(key(), e.getMessage());
```
aneveux, line 81: Log message before maybe?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public String key() {
+        return KEY;
+    }
+
+    @Override
+    public String getDescription() {
+        return "Abstract implementation of GitHub Workflow";
+    }
```
aneveux, line 93: All of this is not needed as it does not make sense to have it in the abstract probe, only in its implementations.

Note that it is not mandatory in an `abstract class` that extends from another `abstract class`, which is your case here, to implement everything coming from the top level class. You can just leave it for the children classes to implement.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
aneveux, line 103: Deserves Javadoc.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
aneveux, line 107: Deserves Javadoc as well.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ContinuousDeliveryProbe.java
```diff
@@ -108,6 +66,6 @@ protected boolean isSourceCodeRelated() {
 
     @Override
     public String[] getProbeResultRequirement() {
-        return new String[] { SCMLinkValidationProbe.KEY, LastCommitDateProbe.KEY };
+        return new String[]{SCMLinkValidationProbe.KEY, LastCommitDateProbe.KEY};
```
aneveux, line 69: Indentation looks weird here, it felt better in the previous version.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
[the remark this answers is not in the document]
aneveux: Makes sense, but it would totally deserve a little bit of documentation =)

If it was me, I would be tempted to create a small utility class for ORDER, add all your explanation as Javadoc, and add some constants to use such as `REQUIRES_PREDECESSOR = 10` or `NEXT_PROBE = 100` for example, and use that here for better readability, like `ORDER = LastCommitDateProbe.ORDER + NEXT_PROBE;`.

WDYT?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
alecharp: to be fair, I used `startsWith` initially to make sure that the version at the end of the string is not a problem and because except the version, it as to be exact.

We could settle for a regexp validation to make sure it's `<workflow-definition>@<version>`.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
alecharp: For Jagruti: you can extract with your IDE the content of the `map` method call into a method. This way you will be able to document it. Make sure to keep it `private` as we don't want it (at least for now) to be re-implemented.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanGithubWorkflowProbe.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public String getFailureMessage() {
+        return "GitHub workflow security scan is not configured in the plugin";
+    }
+
+    @Override
+    public String getSuccessMessage() {
+        return "GitHub workflow security scan is configured in the plugin";
+    }
```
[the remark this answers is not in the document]
Jagrutiti: - I prefer hardcoded strings used in code to be declared final.
- The message in the probes is returned when required in all the other probes. That is why  I followed the same norm. 

I did not notice this until you pointed out :)

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,145 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is an abstract class that looks for desired configuration in the files present in GitHub Workflows directory.
+ *
+ * @return The class returns success when the configuration is found, otherwise returns a failure.
```
aneveux, line 46: There should not be a `@return` in the Javadoc for a class.

As per the general description, it felt a bit too detailed maybe?

How about something like:

```
Abstract Probe allowing to search for the usage of a particular workflow usage within a project's GitHub workflows directory.
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,145 @@
...
+        final Path repository = context.getScmRepository();
+        final Path workflowPath = repository.resolve(WORKFLOWS_DIRECTORY);
+
+        if (!Files.exists(workflowPath)) {
+            return ProbeResult.failure(key(), "Plugin has no GitHub Action configured");
+        }
+
+        try (Stream<Path> files = Files.find(workflowPath, 1, (path, $) -> Files.isRegularFile(path))) {
+            boolean isWorkflowConfigured = files
+                .map(file -> readWorkflowFile(file))
```
aneveux, line 64: `readWorkflowFile` kinda sounds like it's just reading the file, when it is also parsing the Yaml inside and returning it to you if it maps with the object definition you provided.

I feel like the method could be named `parseWorkflowFile` instead, WDYT?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,145 @@
...
+                ProbeResult.success(key(), getSuccessMessage()) :
+                ProbeResult.failure(key(), getFailureMessage());
+        } catch (IOException e) {
+            LOGGER.warn("Couldn't not read file at {} on {}", key(), e);
+            return ProbeResult.error(key(), e.getMessage());
+        }
+    }
+
+    /**
+     * Returns the path to the GitHub Workflow definition which should be use in one of the actions of the plugin repository.
```
aneveux, line 81:
```suggestion
     * Returns the path to the GitHub Workflow definition which should be used in one of the actions of the plugin repository.
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @return the workflow definition used in one of the jobs of one of the actions of the plugin repository.
```
aneveux, line 83: Not exactly, this method returns the path of the workflow you are searching for in the actions defined in the plugin repository. Your sentence sounds like it is actually used, when you don't know yet.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    public abstract String getWorkflowDefinition();
+
+    /**
+     * This method reads the files in GitHub Workflow directory using ObjectMapper
```
aneveux, line 88: It does not read the files, it reads a file, and it parses its Yaml content, and maps it to an object.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ContinuousDeliveryProbe.java
```diff
@@ -24,90 +25,44 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.core.annotation.Order;
 import org.springframework.stereotype.Component;
 
 @Component
 @Order(ContinuousDeliveryProbe.ORDER)
-public class ContinuousDeliveryProbe extends Probe {
-    private static final Logger LOGGER = LoggerFactory.getLogger(ContinuousDeliveryProbe.class);
+public class ContinuousDeliveryProbe extends AbstractGitHubWorkflowProbe {
```
aneveux, line 33: Deserves Javadoc

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,144 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Abstract Probe allows to search for the usage of a particular workflow usage within a project's GitHub workflows directory.
+ */
+
```
aneveux, line 46:
```suggestion
/**
 * Abstract Probe allowing to search for the usage of a particular workflow within a project's GitHub workflows directory.
 */
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * @return the path of the workflow we are searching for.
+     */
+    public abstract String getWorkflowDefinition();
+
+    /**
+     * This method it reads a file, parses its Yaml content, and maps it to an object.
+     *
+     *  @return a partial object mapping of the Yaml content of the file provided in the argument.
+     * */
```
aneveux, line 90:
```suggestion
     */
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,144 @@
...
+            return yaml.readValue(Files.newInputStream(filePath), WorkflowDefinition.class);
+        } catch (IOException e) {
+            LOGGER.warn("Couldn't not read {} for {} on {}", filePath, key(), e);
+            return new WorkflowDefinition(Map.of());
+        }
+    }
+
+    /**
+     * Partial object mapping of a GitHub workflow YAML file, containing only the "jobs" that are defined in it.
+     * */
```
aneveux, line 103:
```suggestion
     */
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
aneveux, line 46: not sure if my suggestion will remove that empty line, but it should go

aneveux:
Just a small bunch of really small indentation/style issues I noticed, but after that the PR should be all good to go, nice job :muscle:

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ContinuousDeliveryProbe.java
alecharp: This seems to be resolved.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanProbe.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+package io.jenkins.pluginhealth.scoring.probes;
+
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(SecurityScanProbe.ORDER)
+public class SecurityScanProbe extends AbstractGitHubWorkflowProbe {
```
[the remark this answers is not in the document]
alecharp: Same.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanProbe.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+/**
+ * This probe checks if Jenkins-infra security scan is configured in the GitHub Actions
+ */
+@Component
+@Order(SecurityScanProbe.ORDER)
+public class SecurityScanProbe extends AbstractGitHubWorkflowProbe {
+    public static final int ORDER = AbstractGitHubWorkflowProbe.ORDER + 10;
```
alecharp, line 36: As the two probes are not dependent to each others, we can use the same order here.

```suggestion
    public static final int ORDER = AbstractGitHubWorkflowProbe.ORDER;
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ContinuousDeliveryProbe.java
```diff
@@ -24,72 +24,18 @@
...
+/**
+ * This probe checks if continuous delivery (CD) workflow is configured in the GitHub Actions
+ */
 @Component
 @Order(ContinuousDeliveryProbe.ORDER)
-public class ContinuousDeliveryProbe extends Probe {
-    private static final Logger LOGGER = LoggerFactory.getLogger(ContinuousDeliveryProbe.class);
-    public static final int ORDER = LastCommitDateProbe.ORDER + 100;
+public class ContinuousDeliveryProbe extends AbstractGitHubWorkflowProbe {
+    public static final int ORDER = AbstractGitHubWorkflowProbe.ORDER + 20;
```
alecharp, line 36: As the two probes are not dependent to each others, we can use the same order here.

```suggestion
    public static final int ORDER = AbstractGitHubWorkflowProbe.ORDER;
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanProbeTest.java
```diff
@@ -0,0 +1,145 @@
...
+            "    uses: jenkins-infra/jenkins-security-scan/.github/workflows/jenkins-security-scan.yaml@v42"
+        ));
+        when(ctx.getScmRepository()).thenReturn(repo);
+
+        final ProbeResult result = probe.apply(plugin, ctx);
+        assertThat(result.status()).isEqualTo(ResultStatus.SUCCESS);
+        assertThat(result.message()).isEqualTo("GitHub workflow security scan is configured in the plugin");
+    }
+}
+
```
alecharp, line 145:
```suggestion
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SecurityScanProbeTest.java
```diff
@@ -0,0 +1,145 @@
...
+import io.jenkins.pluginhealth.scoring.model.ResultStatus;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class SecurityScanProbeTest extends AbstractProbeTest<SecurityScanProbe> {
+
+    Plugin plugin = null;
+    ProbeContext ctx = null;
+    SecurityScanProbe probe = null;
```
alecharp, line 49:
```suggestion
    private Plugin plugin;
    private ProbeContext ctx;
    private SecurityScanProbe probe;
```

alecharp (commented):
This is just nit picking but overall, it looks good. I would like to get it merge today.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,143 @@
...
+     * Returns the path to the GitHub Workflow definition which should be used in one of the actions of the plugin repository.
+     *
+     * @return the path of the workflow we are searching for.
+     */
+    public abstract String getWorkflowDefinition();
+
+    /**
+     * This method it reads a file, parses its Yaml content, and maps it to an object.
+     *
+     *  @return a partial object mapping of the Yaml content of the file provided in the argument.
```
alecharp, line 88:
```suggestion
     * @return a partial object mapping of the Yaml content of the file provided in the argument.
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Checks if the map is null or empty. This means no GitHub action is defined.
+     */
+    private boolean hasWorkflowJobs(WorkflowDefinition workflow) {
+        return workflow.jobs() != null && !workflow.jobs().isEmpty();
+    }
+
+    /**
+     * @return a String array of probes that should be executed before AbstractGitHubWorkflowProbe
+     * */
```
alecharp, line 133:
```suggestion
     */
```

aneveux (approved):

alecharp (approved):

PR #351, opened 2023-07-16
Removed folder path to fix SCMLinkValidationProbe 

### LINKED ISSUE #340: SCMLinkValidationProbe regexp must not have a folder
### Description of the bug

Part of a problem with the `maven-release-plugin`, the plugins stored in a module shouldn't have a `folder`. 
This is why we have to set the `inherite=false` on the `scm` section of the POM files as in https://github.com/jenkinsci/aws-java-sdk-plugin/blob/8b291cfcda0960e5a97fbbd6c9e752049dc3d422/pom.xml#L32.

Because of that, assuming there is a `folder` part in the SCM link of a plugin is wrong. 
https://github.com/jenkins-infra/plugin-health-scoring/blob/8f5da4f2ce554f1c608c63b8942a906c3ccf0459/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java#L47 needs to be fixed and the probe should now fail when there is such a "folder".

Thanks @aneveux

Jagrutiti:
Steps to solve the issue:

1. Remove the `folder` part from Regex expression.
2. Look for `hpi` value in the `<packaging>` tag. 
    - If found, find the value in the `<artifactId>`.
3. If the `<artifactId>` value matches the plugin name
   - Find where the pom lives.
4.  If the pom lives inside the Git repository
    - Store the path to the folder which contains pom file. 
5. Make the folder path available to probe context (that use folders) so that other probes can use the folder and they do not fail.
 
There can be 3 scenarios in the SCM repository:

1. Lonely plugin
2. Plugin code inside module called plugin
3. Bunch of plugins inside modules of git repository

### Description

Closes #340

### Testing done
Did not added new test cases. Modified the existing one.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/HasUnreleasedProductionChangesProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/HasUnreleasedProductionChangesProbe.java
@@ -27,7 +27,6 @@ package io.jenkins.pluginhealth.scoring.probes;
 import static io.jenkins.pluginhealth.scoring.probes.SCMLinkValidationProbe.GH_PATTERN;
 
 import java.io.IOException;
-import java.nio.file.Path;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Comparator;
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/HasUnreleasedProductionChangesProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/HasUnreleasedProductionChangesProbe.java
@@ -27,11 +27,13 @@ package io.jenkins.pluginhealth.scoring.probes;
 import static io.jenkins.pluginhealth.scoring.probes.SCMLinkValidationProbe.GH_PATTERN;
 
 import java.io.IOException;
+import java.nio.file.Path;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.stream.Collectors;
@@ -66,14 +68,14 @@ public class HasUnreleasedProductionChangesProbe extends Probe {
         if (!matcher.find()) {
             return ProbeResult.error(key(), "SCM link doesn't match GitHub plugin repositories");
         }
-        final String folder = context.getScmFolderPath();
+        final Optional<Path> folder = context.getScmFolderPath();
         final Set<String> files = new HashSet<>();
-
         final List<String> paths = new ArrayList<>(3);
         paths.add("pom.xml");
-        if (folder != null) {
-            paths.add(folder + "/pom.xml");
-            paths.add(folder + "/src/main");
+
+        if (folder.isPresent()) {
+            paths.add(folder.get() + "/pom.xml");
+            paths.add(folder.get() + "/src/main");
         } else {
             paths.add("src/main");
         }
@@ -123,6 +125,11 @@ public class HasUnreleasedProductionChangesProbe extends Probe {
     }
 
     @Override
+    public String[] getProbeResultRequirement() {
+        return new String[]{SCMLinkValidationProbe.KEY, LastCommitDateProbe.KEY};
+    }
+
+    @Override
     public String key() {
         return KEY;
     }
@@ -140,10 +147,5 @@ public class HasUnreleasedProductionChangesProbe extends Probe {
          * ProbeEngine, is must be `false`.
          */
         return false;
-    }
-
-    @Override
-    public String[] getProbeResultRequirement() {
-        return new String[]{SCMLinkValidationProbe.KEY, LastCommitDateProbe.KEY};
     }
 }
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/HasUnreleasedProductionChangesProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/HasUnreleasedProductionChangesProbe.java
@@ -66,8 +66,7 @@ public class HasUnreleasedProductionChangesProbe extends Probe {
         if (!matcher.find()) {
             return ProbeResult.error(key(), "SCM link doesn't match GitHub plugin repositories");
         }
-
-        final String folder = matcher.group("folder");
+        final String folder = context.getScmFolderPath();
         final Set<String> files = new HashSet<>();
 
         final List<String> paths = new ArrayList<>(3);
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/HasUnreleasedProductionChangesProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/HasUnreleasedProductionChangesProbe.java
@@ -68,7 +68,7 @@ public class HasUnreleasedProductionChangesProbe extends Probe {
         if (!matcher.find()) {
             return ProbeResult.error(key(), "SCM link doesn't match GitHub plugin repositories");
         }
-        final Optional<Path> folder = context.getScmFolderPath();
+        final Optional<String> folder = context.getScmFolderPath();
         final Set<String> files = new HashSet<>();
         final List<String> paths = new ArrayList<>(3);
         paths.add("pom.xml");
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
@@ -45,6 +45,7 @@ public class ProbeContext {
     private GitHub github;
     private ZonedDateTime lastCommitDate;
     private Map<String, String> pluginDocumentationLinks;
+    private String scmFolderPath;
 
     public ProbeContext(String pluginName, UpdateCenter updateCenter) throws IOException {
         this.updateCenter = updateCenter;
@@ -86,6 +87,14 @@ public class ProbeContext {
     public Optional<String> getRepositoryName(String scm) {
         final Matcher match = SCMLinkValidationProbe.GH_PATTERN.matcher(scm);
         return match.find() ? Optional.of(match.group("repo")) : Optional.empty();
+    }
+
+    public String getScmFolderPath() {
+        return scmFolderPath;
+    }
+
+    public void setScmFolderPath(String scmFolderPath) {
+        this.scmFolderPath = scmFolderPath;
     }
 
     /* default */ void cleanUp() throws IOException {
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
@@ -45,7 +45,7 @@ public class ProbeContext {
     private GitHub github;
     private ZonedDateTime lastCommitDate;
     private Map<String, String> pluginDocumentationLinks;
-    private String scmFolderPath;
+    private Optional<Path> scmFolderPath;
 
     public ProbeContext(String pluginName, UpdateCenter updateCenter) throws IOException {
         this.updateCenter = updateCenter;
@@ -89,11 +89,11 @@ public class ProbeContext {
         return match.find() ? Optional.of(match.group("repo")) : Optional.empty();
     }
 
-    public String getScmFolderPath() {
+    public Optional<Path> getScmFolderPath() {
         return scmFolderPath;
     }
 
-    public void setScmFolderPath(String scmFolderPath) {
+    public void setScmFolderPath(Optional<Path> scmFolderPath) {
         this.scmFolderPath = scmFolderPath;
     }
 
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
@@ -45,7 +45,7 @@ public class ProbeContext {
     private GitHub github;
     private ZonedDateTime lastCommitDate;
     private Map<String, String> pluginDocumentationLinks;
-    private Optional<Path> scmFolderPath;
+    private Optional<String> scmFolderPath;
 
     public ProbeContext(String pluginName, UpdateCenter updateCenter) throws IOException {
         this.updateCenter = updateCenter;
@@ -89,11 +89,11 @@ public class ProbeContext {
         return match.find() ? Optional.of(match.group("repo")) : Optional.empty();
     }
 
-    public Optional<Path> getScmFolderPath() {
+    public Optional<String> getScmFolderPath() {
         return scmFolderPath;
     }
 
-    public void setScmFolderPath(Optional<Path> scmFolderPath) {
+    public void setScmFolderPath(Optional<String> scmFolderPath) {
         this.scmFolderPath = scmFolderPath;
     }
 
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
@@ -97,8 +97,8 @@ public class SCMLinkValidationProbe extends Probe {
         }
         try {
             context.getGitHub().getRepository(matcher.group("repo"));
-            Optional<Path> file = searchPomFiles(context.getScmRepository(), pluginName);
-            String folderPath = file.isPresent() ? new File(file.toString()).getParentFile().getName() : "";
+            Optional<Path> pluginPathInRepository  = findPluginPom(context.getScmRepository(), pluginName);
+            String folderPath = pluginPathInRepository.map(path -> path.getParent().getFileName().toString()).orElse("");
             context.setScmFolderPath(folderPath);
             return ProbeResult.success(key(), "The plugin SCM link is valid.");
         } catch (IOException ex) {
@@ -112,23 +112,22 @@ public class SCMLinkValidationProbe extends Probe {
     }
 
     /**
-     * Searches for Pom files in every directory available in the repository
+     * Searches for Pom file in every directory available in the repository
      *
-     * @param directory  in the scm
+     * @param directory path in the scm
      * @param pluginName the name of the plugin
-     * @return folderPath if it valid or return the scm itself
+     * @return folderPath if it is valid
      */
-    private Optional<Path> searchPomFiles(Path directory, String pluginName) {
-        Optional<Path> fileName = Optional.empty();
+    private Optional<Path> findPluginPom(Path directory, String pluginName) {
         try (Stream<Path> paths = Files.find(directory, 2, (path, $) ->
             path.getFileName().toString().equals("pom.xml"))) {
-            fileName = paths
+            return paths
                 .filter(pom -> pomFileMatchesPlugin(pom, pluginName))
                 .findFirst();
         } catch (IOException e) {
             LOGGER.error("Could not browse the folder during probe {}.", pluginName, e);
         }
-        return fileName;
+        return null;
     }
 
     private boolean pomFileMatchesPlugin(Path pomDirectory, String pluginName) {
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
@@ -24,7 +24,6 @@
 
 package io.jenkins.pluginhealth.scoring.probes;
 
-import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
@@ -97,8 +96,11 @@ public class SCMLinkValidationProbe extends Probe {
         }
         try {
             context.getGitHub().getRepository(matcher.group("repo"));
-            Optional<Path> pluginPathInRepository  = findPluginPom(context.getScmRepository(), pluginName);
-            String folderPath = pluginPathInRepository.map(path -> path.getParent().getFileName().toString()).orElse("");
+            Optional<Path> pluginPathInRepository  = findPluginPom(context.getScmRepository(), pluginName, context.getRepositoryName(scm).get().split("/")[1]);
+            String folderPath = pluginPathInRepository
+                .flatMap(path -> Optional.ofNullable(path.getParent()))
+                .map(parent -> parent.getFileName().toString())
+                .orElseGet(() -> pluginPathInRepository.map(Path::toString).orElse(""));
             context.setScmFolderPath(folderPath);
             return ProbeResult.success(key(), "The plugin SCM link is valid.");
         } catch (IOException ex) {
@@ -118,16 +120,17 @@ public class SCMLinkValidationProbe extends Probe {
      * @param pluginName the name of the plugin
      * @return folderPath if it is valid
      */
-    private Optional<Path> findPluginPom(Path directory, String pluginName) {
-        try (Stream<Path> paths = Files.find(directory, 2, (path, $) ->
+    private Optional<Path> findPluginPom(Path directory, String pluginName, String rootDirectory) {
+        try (Stream<Path> paths = Files.find(directory, Integer.MAX_VALUE, (path, $) ->
             path.getFileName().toString().equals("pom.xml"))) {
             return paths
                 .filter(pom -> pomFileMatchesPlugin(pom, pluginName))
-                .findFirst();
+                .findFirst()
+                .or(() -> Optional.of(Path.of(rootDirectory)));
         } catch (IOException e) {
             LOGGER.error("Could not browse the folder during probe {}.", pluginName, e);
         }
-        return null;
+        return Optional.of(Path.of(rootDirectory));
     }
 
     private boolean pomFileMatchesPlugin(Path pomDirectory, String pluginName) {
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
@@ -68,6 +68,81 @@ public class SCMLinkValidationProbe extends Probe {
         return fromSCMLink(context, plugin.getScm(), plugin.getName());
     }
 
+    /**
+     * Validates the SCM link, and sets the folder path that contains pom.xml.
+     *
+     * @param context    @see {@link ProbeContext}
+     * @param scm        The SCM link @see {@link Plugin#getScm()}
+     * @param pluginName The name of the plugin @see {@link Plugin#getName()}
+     * @return ProbeResult @see {@link ProbeResult}
+     */
+    private ProbeResult fromSCMLink(ProbeContext context, String scm, String pluginName) {
+        Matcher matcher = GH_PATTERN.matcher(scm);
+        if (!matcher.find()) {
+            LOGGER.debug(String.format("%s is not respecting the SCM URL Template.", scm));
+            return ProbeResult.failure(key(), "SCM link doesn't match GitHub plugin repositories.");
+        }
+        try {
+            context.getGitHub().getRepository(matcher.group("repo"));
+            Optional<Path> pluginPathInRepository = findPluginPom(context.getScmRepository(), pluginName);
+            String folderPath = pluginPathInRepository
+                .flatMap(path -> Optional.ofNullable(path.getParent()))
+                .map(parent -> parent.getFileName().toString())
+                .orElse(String.valueOf(context.getScmRepository().getParent().getFileName()));
+            context.setScmFolderPath(folderPath);
+            return ProbeResult.success(key(), "The plugin SCM link is valid.");
+        } catch (IOException ex) {
+            return ProbeResult.failure(key(), "The plugin SCM link is invalid.");
+        }
+    }
+
+    /**
+     * Searches for Pom file in every directory available in the repository
+     *
+     * @param directory  path in the scm
+     * @param pluginName the name of the plugin
+     * @return an Optional path if pom file is found
+     */
+    private Optional<Path> findPluginPom(Path directory, String pluginName) {
+        if (!Files.isDirectory(directory)) {
+            LOGGER.error("Directory {} does not exists during {} probe.", directory, pluginName);
+            return Optional.empty();
+        }
+        /*
+         * If the `maxDepth` is more than 3, we will be navigating the `src/main/java/io/jenkins/plugins/artifactid/` folder
+         * A lot of plugins aren't located deeper than <root>/plugins/<pom.xml>.
+         * */
+        try (Stream<Path> paths = Files.find(directory, 3, (path, $) ->
+            "pom.xml".equals(path.getFileName().toString()))) {
+            return paths
+                .filter(pom -> pomFileMatchesPlugin(pom, pluginName))
+                .findFirst();
+        } catch (IOException e) {
+            LOGGER.error("Could not browse the folder during probe {}. {}", pluginName, e);
+        }
+        return Optional.empty();
+    }
+
+    private boolean pomFileMatchesPlugin(Path pomFilePath, String pluginName) {
+        MavenXpp3Reader mavenReader = new MavenXpp3Reader();
+        try (Reader reader = new InputStreamReader(new FileInputStream(pomFilePath.toFile()), StandardCharsets.UTF_8)) {
+            Model model = mavenReader.read(reader);
+            if ("hpi".equals(model.getPackaging()) && pluginName.equals(model.getArtifactId())) {
+                return true;
+            }
+        } catch (IOException e) {
+            LOGGER.error("Pom file not found for {}.", pluginName, e);
+        } catch (XmlPullParserException e) {
+            LOGGER.error("Could not parse pom file for {}.", pluginName, e);
+        }
+        return false;
+    }
+
+    @Override
+    public String[] getProbeResultRequirement() {
+        return new String[]{UpdateCenterPluginPublicationProbe.KEY};
+    }
+
     @Override
     public String key() {
         return KEY;
@@ -84,72 +159,5 @@ public class SCMLinkValidationProbe extends Probe {
     @Override
     protected boolean requiresRelease() {
         return true;
-    }
-
-    private ProbeResult fromSCMLink(ProbeContext context, String scm, String pluginName) {
-        Matcher matcher = GH_PATTERN.matcher(scm);
-        if (!matcher.find()) {
-            if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug("{} is not respecting the SCM URL Template.", scm);
-            }
-            return ProbeResult.failure(key(), "SCM link doesn't match GitHub plugin repositories.");
-        }
-        try {
-            context.getGitHub().getRepository(matcher.group("repo"));
-            Optional<Path> pluginPathInRepository  = findPluginPom(context.getScmRepository(), pluginName);
-            String folderPath = pluginPathInRepository
-                .flatMap(path -> Optional.ofNullable(path.getParent()))
-                .map(parent -> parent.getFileName().toString())
-                .orElseGet(() -> pluginPathInRepository.map(Path::toString).orElse(""));
-            context.setScmFolderPath(folderPath);
-            return ProbeResult.success(key(), "The plugin SCM link is valid.");
-        } catch (IOException ex) {
-            return ProbeResult.failure(key(), "The plugin SCM link is invalid.");
-        }
-    }
-
-    @Override
-    public String[] getProbeResultRequirement() {
-        return new String[]{UpdateCenterPluginPublicationProbe.KEY};
-    }
-
-    /**
-     * Searches for Pom file in every directory available in the repository
-     *
-     * @param directory path in the scm
-     * @param pluginName the name of the plugin
-     * @return folderPath if it is valid
-     */
-        private Optional<Path> findPluginPom(Path directory, String pluginName) {
-        if (! Files.isDirectory(directory)) {
-            LOGGER.error("Directory {} does not exists during the probe during probe {}.", directory, pluginName);
-            return Optional.empty();
-        }
-
-        try (Stream<Path> paths = Files.find(directory, 3, (path, $) ->
-            "pom.xml".equals(path.getFileName().toString()))) {
-            return paths
-                .filter(pom -> pomFileMatchesPlugin(pom, pluginName))
-                .findFirst()
-                .or(() -> Optional.of(directory.getParent().getFileName()));
-        } catch (IOException e) {
-            LOGGER.error("Could not browse the folder during probe {}.", pluginName, e);
-            return Optional.empty();
-        }
-    }
-
-    private boolean pomFileMatchesPlugin(Path pomFilePath, String pluginName) {
-        MavenXpp3Reader mavenReader = new MavenXpp3Reader();
-        try (Reader reader = new InputStreamReader(new FileInputStream(pomFilePath.toFile()), StandardCharsets.UTF_8)) {
-            Model model = mavenReader.read(reader);
-            if ("hpi".equals(model.getPackaging()) && pluginName.equals(model.getArtifactId())) {
-                return true;
-            }
-        } catch (IOException e) {
-            LOGGER.error("Pom file not found for {}.", pluginName, e);
-        } catch (XmlPullParserException e) {
-            LOGGER.error("Could not parse pom file for {}.", pluginName, e);
-        }
-        return false;
     }
 }
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
@@ -24,7 +24,12 @@
 
 package io.jenkins.pluginhealth.scoring.probes;
 
-import java.io.*;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
@@ -123,11 +128,11 @@ public class SCMLinkValidationProbe extends Probe {
     }
 
     public static String getSCMFolderPath(String filePath, String pluginName) {
-        if(filePath.startsWith("pom.xml")){
+        if (filePath.startsWith("pom.xml")) {
             MavenXpp3Reader mavenReader = new MavenXpp3Reader();
-            try(Reader reader = new InputStreamReader(new FileInputStream(filePath), StandardCharsets.UTF_8)) {
+            try (Reader reader = new InputStreamReader(new FileInputStream(filePath), StandardCharsets.UTF_8)) {
                 Model model = mavenReader.read(reader);
-                if(model.getPackaging().equals("hpi") && model.getArtifactId().equals(pluginName)) {
+                if (model.getPackaging().equals("hpi") && model.getArtifactId().equals(pluginName)) {
                     return filePath + "/pom.xml";
                 }
             } catch (FileNotFoundException e) {
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
@@ -24,6 +24,7 @@
 
 package io.jenkins.pluginhealth.scoring.probes;
 
+import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
@@ -88,7 +89,13 @@ public class SCMLinkValidationProbe extends Probe {
             String folderPath = pluginPathInRepository
                 .flatMap(path -> Optional.ofNullable(path.getParent()))
                 .map(parent -> parent.getFileName().toString())
-                .orElse(String.valueOf(context.getScmRepository().getParent().getFileName()));
+                .orElseGet(() -> {
+                    Path scmRepository = context.getScmRepository();
+                    String defaultFolderName = scmRepository.getParent() != null
+                        ?  scmRepository.getParent().getFileName().toString()
+                        : "";
+                    return defaultFolderName;
+                });
             context.setScmFolderPath(folderPath);
             return ProbeResult.success(key(), "The plugin SCM link is valid.");
         } catch (IOException ex) {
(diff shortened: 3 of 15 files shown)


core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +108,35 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+        if (files != null) {
+            for (File file : files) {
+                if (file.isDirectory()) {
+                    searchPomFiles(file, pluginName);
+                } else {
+                    return getSCMFolderPath(file.getAbsolutePath(), pluginName);
+                }
+            }
+        }
+        return null;
```
aneveux, line 123: How about using https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#find-java.nio.file.Path-int-java.util.function.BiPredicate-java.nio.file.FileVisitOption...-?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +108,35 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+    }
+
+    public static String getSCMFolderPath(String filePath, String pluginName) {
+        if (filePath.startsWith("pom.xml")) {
+            MavenXpp3Reader mavenReader = new MavenXpp3Reader();
+            try (Reader reader = new InputStreamReader(new FileInputStream(filePath), StandardCharsets.UTF_8)) {
+                Model model = mavenReader.read(reader);
+                if (model.getPackaging().equals("hpi") && model.getArtifactId().equals(pluginName)) {
+                    return filePath + "/pom.xml";
```
aneveux, line 132: This is not returning the folder path, but the `pom.xml` path, which is different from what you were retrieving with the previous version of the code.

PierreBtz: I'm not sure we want this code to run on Windows based systems, but I'd recommend not hardcoding path separators and not using String for paths. Rather use the NIO all around and pass paths as `Path` objects, which have methods to resolve path in an OS independent way.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +108,35 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+                }
+            } catch (IOException e) {
+                LOGGER.error("Pom file not found for {}", pluginName);
```
PierreBtz, line 135: Having the pass in the exception message could help debugging an issue.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +108,35 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+                } else {
+                    return getSCMFolderPath(file.getAbsolutePath(), pluginName);
+                }
+            }
+        }
+        return null;
+    }
+
+    public static String getSCMFolderPath(String filePath, String pluginName) {
+        if (filePath.startsWith("pom.xml")) {
```
PierreBtz, line 127: `equals`?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
Jagrutiti: > This is not returning the folder path, but the `pom.xml` path,

So I should return the value of the `artifact id`? Because that is the folder path, right? The name of the repo and the folder?

Jagrutiti: The folder path is everything after the `repo` that leads to? Pom file?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
Jagrutiti: The reason the `HasUnreleasedProductionChangesProbeTest#shouldFailIfThereIsNotReleasedCommitsInModule` is failing because I am not passing the correct folder path.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +111,37 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+            Model model = mavenReader.read(reader);
+            if (model.getPackaging().equals("hpi") && model.getArtifactId().equals(pluginName)) {
+                return filePath;
+            }
+        } catch (IOException e) {
+            LOGGER.error("Pom file not found for {}", pluginName, e);
+        } catch (XmlPullParserException e) {
+            LOGGER.error("Could not parse pom file for {}", pluginName, e);
+        }
+        return scm;
```
Jagrutiti, line 145: Should I return `null` if the file path is not found? Or what should it be?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +111,53 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+     * @param directory in the scm
+     * @param pluginName the name of the plugin
+     * @param scm the valid scm link
+     * @return folderPath from the scm
+     */
+    private static String searchPomFiles(File directory, String pluginName, String scm) {
+        File[] files = directory.listFiles();
+        if (files == null) {
+            return scm;
+        }
```
aneveux, line 127: This should be an error case right?

You are checking for files at the root of the SCM, and there are none, not even a `pom.xml` file?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +111,53 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+        for (File file : files) {
+            try (Stream<Path> paths = Files.find(file.toPath(), 1, (path, $) ->
```
aneveux, line 129: This is not the usage of `Files.find` I had in mind. This function can actually do the recursive discovery of folders, that is what the second argument is for:

`maxDepth - the maximum number of directory levels to search`

You can directly start your function with something like:

`Files.find(directory, SomeMaxValueYouDefine, (path...` and then act on it.

aneveux: And you can also directly search for `pom.xml` files, not directories.

Then you get as a result a list of all found `pom.xml` that you can check with your other function, and return the parent path of whatever matching pom you found.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
Jagrutiti: Here I  first check if the file is a directory, and then pass it here. So if it a directory then it should list files that it contains.

I only look for pom files in the directory and not exactly at the root of the SCM.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
Jagrutiti: Updated the code

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +111,34 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+        MavenXpp3Reader mavenReader = new MavenXpp3Reader();
+
+        try (Stream<Path> paths = Files.find(directory, 1, (path, $) ->
+            path.getFileName().toString().equals("pom.xml")
+        )) {
+            try (Reader reader = new InputStreamReader(new FileInputStream((File) paths), StandardCharsets.UTF_8)) {
+                Model model = mavenReader.read(reader);
+                if (model.getPackaging().equals("hpi") && model.getArtifactId().equals(pluginName)) {
+                    return directory.toString();
+                }
```
aneveux, line 133: In order to ease writing tests, you can extract this to a dedicated function called something like `boolean pomFileMatchesPlugin(Path pomFile, String pluginName)` that will do the `pom.xml` parsing logic and tell you if the `pom.xml` file you provided indeed matches the plugin you are searching for.

Then you can simply use it with `paths.filter(pom -> pomFileMatchesPlugin(pom, pluginName))` which should contain a single entry.

From there, you can retrieve the directory you are searching for.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +111,34 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+     *
+     * @param directory  in the scm
+     * @param pluginName the name of the plugin
+     * @param scm        the valid scm link
+     * @return folderPath if it valid or return the scm itself
+     */
+    private String searchPomFiles(Path directory, String pluginName, String scm) {
+        MavenXpp3Reader mavenReader = new MavenXpp3Reader();
+
+        try (Stream<Path> paths = Files.find(directory, 1, (path, $) ->
```
aneveux, line 126: With a `maxDepth` of `1`, you will just search the root directory, so this is not doing what you want it to do.

Did you write proper tests to check what you wanted to do? Are they passing?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbeTest.java
```diff
@@ -150,4 +152,57 @@ void shouldRecognizeInvalidGitHubUrl() throws Exception {
...
     }
+
+    @Test
+    void shouldReturnCorrectScmFolderPath() throws IOException {
+        final Plugin p1 = mock(Plugin.class);
+        final ProbeContext ctx = mock(ProbeContext.class);
+        final GitHub gh = mock(GitHub.class);
+        final String repositoryName = "jenkinsci/test-repo";
+
+        Path parentDirectory = Files.createTempDirectory("jenkinsci");
```
Jagrutiti, line 163: I have created the temp directory. But in `searchPomFiles` method  throws
`java.nio.file.NoSuchFileException: jenkinsci\test-repo` .

I do not understand why.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +109,38 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+     *
+     * @param directory  in the scm
+     * @param pluginName the name of the plugin
+     * @param scm        the valid scm link
+     * @return folderPath if it valid or return the scm itself
+     */
+    private String searchPomFiles(Path directory, String pluginName, String scm) {
+        try (Stream<Path> paths = Files.find(directory, 2, (path, $) ->
+            path.getFileName().toString().equals("pom.xml")
+        )) {
```
alecharp, line 124: just for readability

```suggestion
        try (Stream<Path> paths = Files.find(directory, 2, (path, $) -> path.getFileName().toString().equals("pom.xml"))) {
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +109,38 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+            LOGGER.error("Could not browse the folder during probe {}", pluginName, e);
+        }
+        return scm;
+    }
+
+    private boolean pomFileMatchesPlugin(Path pomFile, String pluginName) {
+        MavenXpp3Reader mavenReader = new MavenXpp3Reader();
+        try (Reader reader = new InputStreamReader(new FileInputStream(pomFile.toFile()), StandardCharsets.UTF_8)) {
+            Model model = mavenReader.read(reader);
+            if (model.getPackaging().equals("hpi") && model.getArtifactId().equals(pluginName)) {
```
alecharp, line 136: when you do, `object.method()` you can get a `NullPointerException`, even more when you don't control de `object`. Even if in the current case it should be fine, you can protect yourself from this situation by writting 

```suggestion
            if ("hpi".equals(model.getPackaging()) && pluginName.equals(model.getArtifactId()) {
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbeTest.java
```diff
@@ -150,4 +150,30 @@ void shouldRecognizeInvalidGitHubUrl() throws Exception {
...
+        when(ctx.getGitHub()).thenReturn(github);
+        GHRepository repository = ctx.getGitHub().getRepository(repositoryName);
+        when(github.getRepository(repositoryName)).thenReturn(repository);
+
+        when(p1.getName()).thenReturn("test-repo");
+
+        final SCMLinkValidationProbe probe = getSpy();
+        final ProbeResult r1 = probe.apply(p1, ctx);
+
+        assertThat(ctx.getScmFolderPath()).isEqualTo("test-repo");
```
alecharp, line 174: this is incorrect. From what I see, the `scmFolderPath` should be `test-nested-dir-1`

```suggestion
        assertThat(ctx.getScmFolderPath()).isEqualTo("test-nested-dir-2");
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbeTest.java
```diff
@@ -150,4 +150,30 @@ void shouldRecognizeInvalidGitHubUrl() throws Exception {
...
+        final GitHub github = mock(GitHub.class);
+        final String repositoryName = "jenkinsci/test-repo";
+
+        when(p1.getScm()).thenReturn("https://github.com/" + repositoryName);
+        when(p1.getDetails()).thenReturn(Map.of(
+            UpdateCenterPluginPublicationProbe.KEY, ProbeResult.success(UpdateCenterPluginPublicationProbe.KEY, "")
+        ));
+        when(ctx.getGitHub()).thenReturn(github);
+        GHRepository repository = ctx.getGitHub().getRepository(repositoryName);
+        when(github.getRepository(repositoryName)).thenReturn(repository);
```
alecharp, line 167: you are initializing an object with a call to a method that you mock just after the initializing. 
This doesn't seem right. 

If you need to get a `GHRepository` instance, then mock it. 

```suggestion
        GHRepository repository = mock(GHRepository.class);
        when(github.getRepository(repositoryName)).thenReturn(repository);
```

But first, why do you need that instance?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbeTest.java
alecharp: Answered on Gitter but here: `Path#resolve` method is not creating anything. If you need the file or folder to be created, you need to use `Files#createX`.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbeTest.java
Jagrutiti: To mock `            context.getGitHub().getRepository(matcher.group("repo"));` in `fromSCMLink` method

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -118,8 +121,8 @@ public String[] getProbeResultRequirement() {
      * @return folderPath if it valid or return the scm itself
      */
     private String searchPomFiles(Path directory, String pluginName, String scm) {
-        try (Stream<Path> paths = Files.find(directory, 2, (path, $) -> path.getFileName().toString().equals("pom.xml"))) {
-            return paths.filter(pom -> pomFileMatchesPlugin(pom, pluginName)).toString();
+        try (Stream<Path> paths = Files.find(directory.resolve("pom.xml"), 2, (path, $) -> path.getFileName().toString().equals("pom.xml"))) {
+            return paths.filter(pom -> pomFileMatchesPlugin(directory, pluginName)).toString();
```
Jagrutiti, line 125: How do I collect the Paths? It is is of type `Path` and not String. So the `Collectors.toList` and `Collectors.joining` do not work.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -121,8 +121,8 @@ public String[] getProbeResultRequirement() {
      * @return folderPath if it valid or return the scm itself
      */
     private String searchPomFiles(Path directory, String pluginName, String scm) {
-        try (Stream<Path> paths = Files.find(directory.resolve("pom.xml"), 2, (path, $) -> path.getFileName().toString().equals("pom.xml"))) {
-            return paths.filter(pom -> pomFileMatchesPlugin(directory, pluginName)).toString();
+        try (Stream<Path> paths = Files.find(directory, 2, (path, $) -> directory.getFileName().toString().equals("pom.xml"))) {
```
Jagrutiti, line 124: The `directory` in ` Files.find(directory...` never has a file name. I do need at pass it to `pomFileMatchesPlugin` , I resolve it so that it finds the pom file.

Next I want to return `directory.getFileName()` which will be the `folderPath` that I want to return. 

How do I add it collect and return it.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbeTest.java
```diff
@@ -150,4 +151,31 @@ void shouldRecognizeInvalidGitHubUrl() throws Exception {
...
+        when(ctx.getGitHub()).thenReturn(github);
+        GHRepository repository = mock(GHRepository.class);
+        when(github.getRepository(repositoryName)).thenReturn(repository);
+
+        when(p1.getName()).thenReturn("test-repo");
+
+        final SCMLinkValidationProbe probe = getSpy();
+        final ProbeResult r1 = probe.apply(p1, ctx);
+
+        assertThat(ctx.getScmFolderPath()).isEqualTo("src/test/resources/jenkinsci/test-repo/test-nested-dir-1/test-nested-dir-2");
```
alecharp, line 176: Not the correct value. Because the repository is `src/test/resources/jenkinsci/test-repo/test-nested-dir-1`, the plugin folder path should be `test-nested-dir-2`. If the test fails because of this, it means you need to fix your code, not the other way around.

```suggestion
        assertThat(ctx.getScmFolderPath()).isEqualTo("test-nested-dir-2");
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -97,4 +107,37 @@ private ProbeResult fromSCMLink(ProbeContext context, String scm) {
...
+     *
+     * @param directory  in the scm
+     * @param pluginName the name of the plugin
+     * @param scm        the valid scm link
+     * @return folderPath if it valid or return the scm itself
+     */
+    private String searchPomFiles(Path directory, String pluginName, String scm) {
+        try (Stream<Path> paths = Files.find(directory, 2, (path, $) ->
+            Files.isRegularFile(path.resolve("pom.xml")) && path.getFileName().toString().equals("pom.xml"))) {
+            return paths.filter(pom -> pomFileMatchesPlugin(pom, pluginName)).toString();
```
alecharp, line 122: You could use `findFirst` method on the Stream.

alecharp (changes requested):

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbeTest.java
```diff
@@ -148,6 +153,60 @@ void shouldRecognizeInvalidGitHubUrl() throws Exception {
...
-        assertThat(r1.message()).isEqualTo("The plugin SCM link is invalid");
+        assertThat(r1.message()).isEqualTo("The plugin SCM link is invalid.");
+    }
+
+    @Test
+    void shouldReturnCorrectScmFolderPath() throws IOException {
+        final Plugin plugin = mock(Plugin.class);
+        final GitHub github = mock(GitHub.class);
+        final String repositoryName = "jenkinsci/test-repo";
+        ProbeContext ctx = new ProbeContext(plugin.getName(), new UpdateCenter(Map.of(), Map.of(), List.of()));
```
alecharp, line 164: Why do you need this one?

alecharp (changes requested):

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbeTest.java
Jagrutiti: Do you mean this line?

>         ProbeContext ctx = new ProbeContext(plugin.getName(), new UpdateCenter(Map.of(), Map.of(), List.of()));

I initliaized this to use `spy`.

Do you suggest I do it the following way:

`ProbeContext contextSpy = spy(new ProbeContext(plugin.getName(), new UpdateCenter(Map.of(), Map.of(), List.of())));`?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +87,62 @@ protected boolean requiresRelease() {
...
+                LOGGER.debug("{} is not respecting the SCM URL Template.", scm);
             }
-            return ProbeResult.failure(key(), "SCM link doesn't match GitHub plugin repositories");
+            return ProbeResult.failure(key(), "SCM link doesn't match GitHub plugin repositories.");
         }
-
         try {
             context.getGitHub().getRepository(matcher.group("repo"));
-            return ProbeResult.success(key(), "The plugin SCM link is valid");
+            Optional<Path> file = searchPomFiles(context.getScmRepository(), pluginName);
```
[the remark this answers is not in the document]
aneveux: Also, `file` is not a very descriptive name for the variable. It is something like `pathInRepository` or `pluginPath` or `pluginPathInRepository` or something like this? WDYT?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +87,62 @@ protected boolean requiresRelease() {
...
+
+    /**
+     * Searches for Pom files in every directory available in the repository
+     *
+     * @param directory  in the scm
+     * @param pluginName the name of the plugin
+     * @return folderPath if it valid or return the scm itself
+     */
+    private Optional<Path> searchPomFiles(Path directory, String pluginName) {
+        Optional<Path> fileName = Optional.empty();
```
aneveux, line 122: Why do you need to initialize an `Optional` to empty?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +87,62 @@ protected boolean requiresRelease() {
...
+        try (Stream<Path> paths = Files.find(directory, 2, (path, $) ->
+            path.getFileName().toString().equals("pom.xml"))) {
+            fileName = paths
```
aneveux, line 125: You can directly 

`return directories.filter(pom -> pomFileMatchesPlugin(pom, pluginName)).findFirst();`

instead of having that assigned to variables. If you read the docs of `findFirst`, it tells you that it'll return an empty `Optional` if nothing is found, there is no need to repeat that yourself.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +87,62 @@ protected boolean requiresRelease() {
...
             }
-            return ProbeResult.failure(key(), "SCM link doesn't match GitHub plugin repositories");
+            return ProbeResult.failure(key(), "SCM link doesn't match GitHub plugin repositories.");
         }
-
         try {
             context.getGitHub().getRepository(matcher.group("repo"));
-            return ProbeResult.success(key(), "The plugin SCM link is valid");
+            Optional<Path> file = searchPomFiles(context.getScmRepository(), pluginName);
+            String folderPath = file.isPresent() ? new File(file.toString()).getParentFile().getName() : "";
```
aneveux, line 101: Am I properly understanding that in your code, it is valid for a repository to NOT contain a `pom.xml` for a plugin?

AFAICT, you SHOULD always find a `pom.xml` matching your plugin in the repository. Either it is at the root of the repository or not, but it should exist.

If it is not found, it means that the plugin you are searching for does not exist in your repository, and I feel like this Probe is not dealing with that situation.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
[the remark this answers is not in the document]
Jagrutiti: Thank you for pointing these things out. It was a local variable in a method so I didn't put much thought into naming it.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +87,62 @@ protected boolean requiresRelease() {
...
 
     @Override
     public String[] getProbeResultRequirement() {
         return new String[]{UpdateCenterPluginPublicationProbe.KEY};
     }
+
+    /**
+     * Searches for Pom files in every directory available in the repository
+     *
+     * @param directory  in the scm
```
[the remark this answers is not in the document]
Jagrutiti: Fixed it to `directory path in the scm`.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
Jagrutiti: Because I cannot return the variable after `try` statement if the variable is not initialized. 

The other option is that I initialize it to `null`.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +87,62 @@ protected boolean requiresRelease() {
...
+     * @param pluginName the name of the plugin
+     * @return folderPath if it valid or return the scm itself
+     */
+    private Optional<Path> searchPomFiles(Path directory, String pluginName) {
+        Optional<Path> fileName = Optional.empty();
+        try (Stream<Path> paths = Files.find(directory, 2, (path, $) ->
```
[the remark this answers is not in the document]
Jagrutiti: You are right about this. 

I discussed with Adrien in our last weekly meeting. And he said `2` is the ideal depth. He also added that we can go till 5 but it would be a little crazy number.

I was thinking maybe we should find the max depth of a directory and then set it accordingly.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
Jagrutiti: I read the next comment. Which was related to this one and fixed it.

Jagrutiti: I had this doubt of what to return after `try` statement. 

For now I am returning `null`.

aneveux: Do not return `null`, return `Optional.empty()`. You just don't need variables being assigned here.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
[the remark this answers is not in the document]
aneveux: I think whatever the value is, it should be clear why we chose that value.

I'd go for either

- Use Integer.MAX_VALUE so we scan everything
- Use a fixed value that we consider the maximum supported/recommended value and we document it as such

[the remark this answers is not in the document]
Jagrutiti: I went for Integer.MAX_VALUE

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
alecharp: > AFAICT, you SHOULD always find a pom.xml matching your plugin in the repository. Either it is at the root of the repository or not, but it should exist.

we still have 80 plugins using Gradle as build tool.

The problem with gradle is that it's really complicated to parse the build description.

aneveux: Oooooh that is right! I totally forgot about that, thanks for the reminder @alecharp!

How do you consider those plugins for the other probes requiring Maven stuff btw?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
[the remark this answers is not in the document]
aneveux: The rationale looks good to me, I just think it needs to be explicit and clear in the source code.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
alecharp: For the moment, with difficulties. It's a problem we know (we probably should advertise that a bit more).

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +86,65 @@ protected boolean requiresRelease() {
         return true;
     }
 
-    private ProbeResult fromSCMLink(ProbeContext context, String scm) {
+    private ProbeResult fromSCMLink(ProbeContext context, String scm, String pluginName) {
         Matcher matcher = GH_PATTERN.matcher(scm);
         if (!matcher.find()) {
             if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug("{} is not respecting the SCM URL Template", scm);
```
alecharp, line 83: This is just a log, it won't be displayed to users. That's why it's not a problem to not have a period at the end of the sentence. 
But don't revert your change.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +86,65 @@ protected boolean requiresRelease() {
...
+
+    /**
+     * Searches for Pom file in every directory available in the repository
+     *
+     * @param directory path in the scm
+     * @param pluginName the name of the plugin
+     * @return folderPath if it is valid
+     */
+    private Optional<Path> findPluginPom(Path directory, String pluginName, String rootDirectory) {
+        try (Stream<Path> paths = Files.find(directory, Integer.MAX_VALUE, (path, $) ->
```
alecharp, line 124:
```suggestion
        try (Stream<Path> paths = Files.find(directory, 3, (path, $) ->
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +86,65 @@ protected boolean requiresRelease() {
...
+            path.getFileName().toString().equals("pom.xml"))) {
```
alecharp, line 125: Please use the `"".equals(object.method())` format here.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +86,65 @@ protected boolean requiresRelease() {
...
+            return paths
+                .filter(pom -> pomFileMatchesPlugin(pom, pluginName))
+                .findFirst()
+                .or(() -> Optional.of(Path.of(rootDirectory)));
+        } catch (IOException e) {
+            LOGGER.error("Could not browse the folder during probe {}.", pluginName, e);
```
alecharp, line 131: If we get an exception, it probably means that we couldn't validate the folder, which could lead to more issues later on. 
Please return `Optional.empty()` here and make sure to handle that case in the method.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +86,65 @@ protected boolean requiresRelease() {
...
+        }
+        return Optional.of(Path.of(rootDirectory));
+    }
+
+    private boolean pomFileMatchesPlugin(Path pomDirectory, String pluginName) {
```
alecharp, line 136: The first parameter is a file or a directory? It seems to be a file, but it's name suggest the opposite. Please fix that.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +86,65 @@ protected boolean requiresRelease() {
...
     }
+
+    /**
+     * Searches for Pom file in every directory available in the repository
+     *
+     * @param directory path in the scm
+     * @param pluginName the name of the plugin
+     * @return folderPath if it is valid
+     */
+    private Optional<Path> findPluginPom(Path directory, String pluginName, String rootDirectory) {
```
alecharp, line 123: Why do you need the third parameter? Just to return it if something wrong is happening? Please don't, handle this in the calling method.

alecharp (changes requested):

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +86,70 @@ protected boolean requiresRelease() {
...
+            LOGGER.error("Directory {} does not exists during the probe during probe {}.", directory, pluginName);
+            return Optional.empty();
+        }
+
+        try (Stream<Path> paths = Files.find(directory, 3, (path, $) ->
+            "pom.xml".equals(path.getFileName().toString()))) {
+            return paths
+                .filter(pom -> pomFileMatchesPlugin(pom, pluginName))
+                .findFirst()
+                .or(() -> Optional.of(directory.getParent().getFileName()));
```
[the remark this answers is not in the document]
Jagrutiti: In the previous reviews, you said that a `pom` file should always exist in the Maven plugin. Which is true. 

I assumed a pom file exist in the default parent directory. 

> Finding the path of the parent directory, either of the found pom.xml or the directory you were searching in, should be done elsewhere.

Do you mean when the `Optional.empty` is returned I should write the logic to look for its parent in another method?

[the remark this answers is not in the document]
aneveux: What I mean is that a function should return a consistent result.

Here, you either return a path that ends with `/pom.xml` if it is found, or you return a path that does not end by `/pom.xml` if the file is not found.

What I am saying is that it should be clear and consistent:

- Either you return the path ending with `/pom.xml` if you find it, and `empty` if you don't find the file (which you are right I said should be an error)
- Or you return the path that is supposed to contain the `pom.xml` inferring what it should be as a default case if the file is not found.

A mix of approaches does not make the code very clear.

I would personally advise to return the `/pom.xml` path if the file is found, and throw an error if the file is not found. The only case in which this will happen (according to Adrien) is for Gradle plugins, but I assume it's OK that the probe is broken in that case.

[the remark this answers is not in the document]
Jagrutiti: I understand. You want me the method to do one thing at a time for better readability. 

I have followed your advise. Thanks.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -76,25 +86,70 @@ protected boolean requiresRelease() {
         return true;
     }
 
-    private ProbeResult fromSCMLink(ProbeContext context, String scm) {
+    private ProbeResult fromSCMLink(ProbeContext context, String scm, String pluginName) {
         Matcher matcher = GH_PATTERN.matcher(scm);
         if (!matcher.find()) {
             if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug("{} is not respecting the SCM URL Template", scm);
+                LOGGER.debug("{} is not respecting the SCM URL Template.", scm);
```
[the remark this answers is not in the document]
Jagrutiti: I will work on this. I am getting the syntax incorrect for `debug`. Because it does not take these parameters.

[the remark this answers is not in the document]
Jagrutiti: Hey @aneveux 

I could not find an example with Lambda in the documentation. I came across this link: https://www.slf4j.org/faq.html#logging_performance

This speaks of the benefits and overheads of the current and the previous syntax. 

In the[ documentation](https://www.slf4j.org/apidocs/org/slf4j/Logger.html#debug(java.lang.String%2C%20java.lang.Object...)) where different signatures are explained, there is no method with for lambda syntax. That was the error I was getting yesterday.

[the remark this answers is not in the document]
aneveux: Ooooh my bad... I'm too used of relying on Java logging, not SLF4J.

In SLF4J that approach exists, but it uses their fluent API, so using something like `LOGGER.atDebug` with proper arguments. In there, you can provide a Supplier, and the `atDebug` (or whatever `atSomething` method) will get called only if the LOGGER is enabled for that specific level.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -42,20 +53,133 @@
...
+                .flatMap(path -> Optional.ofNullable(path.getParent()))
+                .map(parent -> parent.getFileName())
+                .orElseGet(() -> {
+                    Path scmRepository = context.getScmRepository();
+                    Optional<Path> pluginPomInRootDirectory = findPluginPomInRootDirectory(context.getScmRepository(), pluginName);
+                    Path defaultFolder = pluginPomInRootDirectory.isPresent()
+                        ? scmRepository.getParent().getFileName()
+                        : Path.of("");
+                    return defaultFolder;
+                });
```
aneveux, line 103: I feel there is something you misunderstood about the file searching:

`Files.find(directory, 3, (path, $)` means that it will recursively search for the `pom.xml` file recursively, from the root directory (`directory`) up to 3 nested levels.

Meaning that if your plugin's `pom.xml` is located at the root of the repository, it will be found by `findPluginPom` directly.

There is no need for `findPluginPomInRootDirectory` as it won't catch anything else.

Your new implementation of `findPluginPom` is looking good to me. It is searching for the pom path, and if it is not found, it will return an empty optional, which is what we want.

Now for retrieving `folderPath`, what you want is:

*  If a `pom.xml` was found, the directory is its parent (cause the path ends with `/pom.xml` so we want the folder containing it)
* If no `pom.xml` was found, it is an error.

So, a simple:

```
Optional<Path> folderPath = pluginPathInRepository.map(path -> path.getParent());
```

If `folderPath` is empty, it means no `pom.xml` was found, it is an error (it will fail for Gradle plugins, but that's OK).

if `folderPath` is not empty, it will contain the path to the directory containing the `pom.xml`, which is what we want. We can use that in the `context` and return the proper result.

Jagrutiti: One of the test cases was incorrect. And to fix that I messed up with the code.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
```diff
@@ -88,6 +89,14 @@ public Optional<String> getRepositoryName(String scm) {
         return match.find() ? Optional.of(match.group("repo")) : Optional.empty();
     }
 
+    public Optional<Path> getScmFolderPath() {
```
Jagrutiti, line 92: I changed the return type to `Optional<Path>` I hope it is okay. Adrien said in the meeting he wants it to be `String`. Would this make a big difference? 

Also, it aligns with the method name like `getScmFolderPath` returns a `Path` which is one of the things you always tell me to focus on: naming methods and conventions.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/HasUnreleasedProductionChangesProbe.java
```diff
@@ -66,15 +68,14 @@ public ProbeResult doApply(Plugin plugin, ProbeContext context) {
         if (!matcher.find()) {
             return ProbeResult.error(key(), "SCM link doesn't match GitHub plugin repositories");
         }
-
-        final String folder = matcher.group("folder");
+        final Optional<Path> folder = context.getScmFolderPath();
```
aneveux, line 71: I am not entirely convinced we can really store the folder path as a `Path`. Did you make sure that with this implementation the `Path` object does contain only the folder path and not some root element as well?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
aneveux: Yeah at first (without playing with it a bit to check edge cases), I would be tempted to say it's better to keep the object as a `String`. 

We really do just want to get the folders that are there, and a `Path` object can contain more.

The fact that the name of the field is `ScmFolderPath` is because we indeed return the folder path, it should not have something to do with the Java type `Path`. Introducing the object type in the variable name is not that great of a practice, even though it was done for years.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -42,20 +53,94 @@
...
+     * @return ProbeResult {@link ProbeResult}.
+     */
+    private ProbeResult fromSCMLink(ProbeContext context, String scm, String pluginName) {
+        Matcher matcher = GH_PATTERN.matcher(scm);
+        if (!matcher.find()) {
+            LOGGER.atDebug().log(() -> String.format("%s is not respecting the SCM URL Template.", scm));
+            return ProbeResult.failure(key(), "SCM link doesn't match GitHub plugin repositories.");
+        }
+        try {
+            context.getGitHub().getRepository(matcher.group("repo"));
```
aneveux, line 86: That line is weird, cause a `get` method, often called a `getter`, is expected to return something, while here I guess it is just used in order to clone the repository right?

I don't expect you to change the code of the context object from there, that would be another refactoring, but a one line comment saying that this line clones the repository allows to understand why it is there.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -42,20 +53,94 @@
...
+            Optional<Path> pluginPathInRepository = findPluginPom(context.getScmRepository(), pluginName);
+            Optional<Path> folderPath = pluginPathInRepository.map(path -> path.getParent());
+            if (folderPath.isEmpty()) {
+                return ProbeResult.error(key(), String.format("No valid POM file found in %s plugin.", pluginName));
+            }
+            context.setScmFolderPath(Optional.of(folderPath.get().getFileName()));
```
aneveux, line 92: Here I believe it would be better to have the ScmFolderPath being a `Optional<String>`.

Considering `folderPath` is already an `Optional`, you can simply go for something like:

`context.setScmFolderPath(folderPath.map( whatever function gives you the value you expected in there ));`

If the `Optional` was empty, the map function won't be running, and you'll store an empty again. Otherwise, the Optional will contain a `String` with the value you wanted.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
```diff
@@ -42,20 +53,94 @@
...
+            return paths
+                .filter(pom -> pomFileMatchesPlugin(pom, pluginName))
+                .findFirst();
+        } catch (IOException e) {
+            LOGGER.error("Could not browse the folder during probe {}. {}", pluginName, e);
+        }
+        return Optional.empty();
+    }
+
+    private boolean pomFileMatchesPlugin(Path pomFilePath, String pluginName) {
```
aneveux, line 126: Could deserve some Javadocs.

aneveux (commented):
Relatively minor comments, but that is a great job here. Well done Jagruti.

Some general remarks
- You should take the habit of writing Javadocs for everything
- When designing data for tests, you could go for names which better express the expected result, or the intent.

About test data for example: instead of test-nested-dir-2 you could write something like `expectedNestedDirectory` or something like this. And for something that fails, something like `thisNestedDirectoryShouldNotBeFound` or something very clear in that sense. No need to modify the current tests though, it's just a general remark.

Jagrutiti:
I will make a habit of writing more JavaDocs and giving meaningful names to the test resources.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/SCMLinkValidationProbe.java
aneveux: Looks good the way you wrote it in [6f90c9b](https://github.com/jenkins-infra/plugin-health-scoring/pull/351/commits/6f90c9b255cf7f26faafec3dae12d253e17f4eed)

aneveux (approved):
LGTM. I'll let @alecharp do another review of it just to be sure I did not miss anything, but I believe it should be ready to merge 🚀

alecharp (approved):

PR #352, opened 2023-07-18
Fixes logger message format

alecharp:
### Description

Follow up of #342.

This aims to fix the logger messages format when the probe is not able to parse the files in the plugin Git repository.
I also used method references to simplify the code syntax.

### Testing done

Only syntax changes, no new code.

```[tasklist]
### Submitter checklist

- [ ] If an issue exists, it is well described and linked in the description
```

### DIFF
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
@@ -59,8 +59,8 @@ public abstract class AbstractGitHubWorkflowProbe extends Probe {
 
         try (Stream<Path> files = Files.find(workflowPath, 1, (path, $) -> Files.isRegularFile(path))) {
             boolean isWorkflowConfigured = files
-                .map(file -> parseWorkflowFile(file))
-                .filter(workflow -> hasWorkflowJobs(workflow))
+                .map(this::parseWorkflowFile)
+                .filter(this::hasWorkflowJobs)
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
```diff
@@ -59,8 +59,8 @@ protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
 
         try (Stream<Path> files = Files.find(workflowPath, 1, (path, $) -> Files.isRegularFile(path))) {
             boolean isWorkflowConfigured = files
-                .map(file -> parseWorkflowFile(file))
-                .filter(workflow -> hasWorkflowJobs(workflow))
+                .map(this::parseWorkflowFile)
```
Jagrutiti, line 62: When you use `this::parseWorkflowFile` this means the code will look for `parseWorkflowFile` in the same class?

and the parameter will be passed on its own? We do not need to use the arrow operator?

Jagrutiti (approved):

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractGitHubWorkflowProbe.java
alecharp: I could try to explain to you, but there is an excellent article about that method reference: https://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html

PR #353, opened 2023-07-23
Number of open issue probe

### LINKED ISSUE #143: Number of open issues
Having the number of issues for a plugin can help understand the interaction between the users and the maintainers.

Jagrutiti:
### Description

This probe fetches the number of open issues in a plugin. 

Steps:

Create `IssueTrackerDetectionProbe`, this probe will check whether JIRA, GitHub is configured or both is configured. 
- Return a success as long as at least one issue tracker configured.
- If no issue tracker is configured that is an error
- more than one error is a success. Stores which ones are configured.

For  `GitHubOpenIssuesProbe` and `JiraOpenIssuesProbe`:

- Create an abstract class. That would be the parent for `GitHubOpenIssuesProbe` and `JiraOpenIssuesProbe`.
- Make sure that `IssueTrackerDetectionProbe`  is executed and success is returned.
-  Get the message of the probe and pass it. Try to see which system is used. 
- Return a `Map<key, url>`

Some code reusing details:
> Basically this from your PR:

 ```   private List<String> getIssueTrackerData(ProbeContext context, String filter) {
        return context.getUpdateCenter()
            .issueTrackers().stream()
            .flatMap(map -> map.entrySet().stream())
            .filter(map -> map.getKey().equals(filter))
            .map(map -> map.getValue())
            .collect(Collectors.toList());
    }
```
Is your entry point for the new probe
```
IssueTrackerDetectionProbe
| - GitHubOpenIssuesProbe
| - JiraOpemIssuesProbe
```

```ProbeContext.getIssuesTracker() ->
"github" - "https://github.com/
"jira" - "https://issues.jenkins.io
```

Closes #143 

### Testing done

```[tasklist]
### Submitter checklist
```


core/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
alecharp, line 1: No. Don't change the model that is stored in Database with intermediate data. 
The fact that you need to have the issue tracker data is only for the new probe.

alecharp (changes requested):

core/src/main/java/io/jenkins/pluginhealth/scoring/model/Plugin.java
Jagrutiti: I have removed it from the models and added it to the record `UpdateCenter`. 

Test cases that mock the `UpdateCenter` are failing because they do not have the new parameter.

Is this the desired solution. Shall I update the other test cases?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/NumberOfOpenIssuesProbeTest.java
```diff
@@ -0,0 +1,235 @@
...
+        when(restTemplate.getForEntity(anyString(), anyString().getClass())).thenReturn(responseEntity);
+        when(responseEntity.getBody()).thenReturn(jsonString);
+        when(jsonNode.get(anyString())).thenReturn(new ObjectMapper().readTree(jsonString));
+
+        final NumberOfOpenIssuesProbe probe = getSpy();
+
+        assertThat(probe.apply(plugin, ctx))
+            .usingRecursiveComparison()
+            .comparingOnlyFields("id", "status", "message")
+            .isEqualTo(ProbeResult.success(NumberOfOpenIssuesProbe.KEY, "67 open issues found in JIRA."));
```
Jagrutiti, line 179: Since this is a live plugin. I wanted to mock the total number `67`.

But I wasn't able to to. From the debugger I could make a guess the API call in the code overrides the mock. But am not sure.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/NumberOfOpenIssuesProbeTest.java
```diff
@@ -0,0 +1,235 @@
...
+                    "github",
+                    "viewUrl",
+                    "https://github.com/" + repository + "/issues"
+                )
+            )
+        ));
+        when(ctx.getGitHub()).thenReturn(gh);
+        when(ctx.getRepositoryName(plugin.getScm())).thenReturn(Optional.of(repository));
+        when(gh.getRepository(repository)).thenReturn(ghRepository);
+        when(ghRepository.getOpenIssueCount()).thenReturn(6);
```
Jagrutiti, line 221: I am not sure if mocking `getOpenIssueCount` is the right thing to do. 

Because this actually being test. I took the inspiration from `CodeCoverageProbe` where most of the where functions used from the GitHub library were mocked.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/NumberOfOpenIssuesProbeTest.java
```diff
@@ -0,0 +1,235 @@
...
+            .usingRecursiveComparison()
+            .comparingOnlyFields("id", "status", "message")
+            .isEqualTo(ProbeResult.success(NumberOfOpenIssuesProbe.KEY, "6 open issues found in GitHub."));
+        verify(probe, atMostOnce()).doApply(plugin, ctx);
+
+    }
+}
+
+
+
```
alecharp, line 235:
```suggestion
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/NumberOfOpenIssuesProbeTest.java
```diff
@@ -0,0 +1,235 @@
...
+        when(gh.getRepository(repository)).thenReturn(ghRepository);
+        when(ghRepository.getOpenIssueCount()).thenReturn(6);
+
+        final NumberOfOpenIssuesProbe probe = getSpy();
+
+        assertThat(probe.apply(plugin, ctx))
+            .usingRecursiveComparison()
+            .comparingOnlyFields("id", "status", "message")
+            .isEqualTo(ProbeResult.success(NumberOfOpenIssuesProbe.KEY, "6 open issues found in GitHub."));
+        verify(probe, atMostOnce()).doApply(plugin, ctx);
```
alecharp, line 229: What does `atMostOnce` method does?

```suggestion
        verify(probe).doApply(plugin, ctx);
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/NumberOfOpenIssuesProbeTest.java
alecharp: What you mock is what you can control but is not in the project. Does `GhRepository` class fit the description? Are you testing its method `getOpenIssueCount` behavior?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/NumberOfOpenIssuesProbe.java
```diff
@@ -0,0 +1,150 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(NumberOfOpenIssuesProbe.class);
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        List<String> issueTracker = getIssueTrackerType(context);
+
+        if (issueTracker.size() > 1) {
+            return ProbeResult.success(key(),
+                getJiraIssues(context, plugin.getScm(), plugin.getName()).message().concat(" ") +
+                getGitHubIssues(context, plugin.getScm(), plugin.getName()).message());
```
alecharp, line 66: You are concatenating two `ProbeResult`. How does that work?

alecharp (changes requested):

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/NumberOfOpenIssuesProbe.java
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * This probe counts the total number of open issues in GitHub and JIRA
+ */
+@Component
+@Order(value = NumberOfOpenIssuesProbe.ORDER)
+public class NumberOfOpenIssuesProbe extends Probe {
+    public static final String KEY = "open-issue";
+    public static final int ORDER = UpdateCenterPluginPublicationProbe.ORDER + 100;
+    public static final String JIRA_HOST = "https://issues.jenkins.io/";
+    public static final String JIRA_API_PATH = "rest/api/latest/search";
```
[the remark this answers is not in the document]
Jagrutiti: The API path might change in future or may have queryStrings.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/NumberOfOpenIssuesProbe.java
Jagrutiti: When there are open issues in both `JIRA` and `GitHub` both,  both methods will return the count of open issues. 

Their message is then combined and returned to the user. 

I tried using a switch case and also played around with lambda. But it did not work.

alecharp: you are combining the `toString()` of both `ProbeResult`. If that method is changed to not only returns the content of the `message` field, this won't work anymore. 

Moreover, what happens if `getJiraIssues` is returning an error?

Jagrutiti: Yes. I was thinking the same. It will return a failure message. 

I want to use these two methods. And resolve the case when there are open issues in both the tracker type.

core/src/main/java/io/jenkins/pluginhealth/scoring/model/updatecenter/UpdateCenter.java
```diff
@@ -29,6 +29,8 @@
 
 public record UpdateCenter(Map<String, Plugin> plugins,
                            Map<String, Deprecation> deprecations,
-                           List<SecurityWarning> warnings
+                           List<SecurityWarning> warnings,
+                           List<Map<String, String>> issueTrackers
+
```
alecharp, line 34: This does not exists.
Please see https://updates.jenkins.io/current/update-center.actual.json.

The issue trackers are listed within each plugin.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbe.java
```diff
@@ -0,0 +1,69 @@
...
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        Map<String, String> issueTrackerType = context.getUpdateCenter()
+            .issueTrackers().stream()
+            .collect(Collectors.toMap(
+                issueTracker -> issueTracker.get("type"),
+                issueTracker -> issueTracker.get("viewUrl")
+            ));
+
+        context.setIssueTrack(issueTrackerType);
+        return ProbeResult.success(key(), "Issue tracker detected and returned successfully.");
```
alecharp, line 52: See https://github.com/jenkins-infra/plugin-health-scoring/pull/353/files#r1283269733. 
Because the issue trackers list is within each plugin, you need to check if the plugin is inside the update-center, then you can try to find its issue trackers list and register them.

core/src/main/java/io/jenkins/pluginhealth/scoring/model/updatecenter/UpdateCenter.java
Jagrutiti: For this I need to call the `UpdateCenterPluginPublicationProbe`? 

if it succeeds go ahead?

Jagrutiti: And remove this `List<Map<String, String>> issueTrackers`?

alecharp: you need to get the plugin from the update-center list and from it get its issue trackers list.

core/pom.xml
```diff
@@ -102,5 +102,9 @@
 			<artifactId>postgresql</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-web</artifactId>
+		</dependency>
```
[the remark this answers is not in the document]
Jagrutiti: `spring-boot-starter-web` adds 4 different dependencies. While `spring-web` adds only web dependencies. 

So added the latter.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractOpenIssuesProbe.java
```diff
@@ -0,0 +1,137 @@
...
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.kohsuke.github.GHRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.RestTemplate;
+
+public abstract class AbstractOpenIssuesProbe extends Probe {
```
aneveux, line 48: Hey @Jagrutiti,

There is something wrong with that class which comes from its design.

On purpose, I won't give you more indication for now, to let you figure it out by yourself.

Please have a look at other similar classes in the project, and please read more about the purpose of `abstract` classes, and the general purpose of abstraction in OOP. 

You already implemented similar kind of classes in previous PRs, but perhaps the extra guidance we provided made that you did not get to understand the reason why we recommended this design, which is why this time I won't share much details (yet) and let you take some time to think about the design choices and understand the reason for them.

Jagrutiti: Hi @aneveux ,

I did read about implementing an `abstract` class in OOPS. 

The articles spoke of data abstraction and functional abstraction. In our case I believe, we are doing functional abstraction. 

Having two different methods is important here because the implementation to count issues from both `JIRA` and `GitHub`  have nothing in common.

I had done a similar implementation in `AbstractDependencyBotConfigurationProbe` but in that case `RenovateProbe` and `DependabotProbe` had the same job. Look for files only with different names.

The next point that I come across is why `abstract` classes are used. They are designed for:
-  hide complex functionality
- code reusability

The abstract classes that I implemented did provide code reusability and their functionalities were hidden from the child class. They only had to implement the abstract class.

In the `NumberofOpenIssuesProbe` code reusability is not possible because the libraries and API used in both cases are different. 

About the design issues, `JIRA_HOST` and `restTemplate` are used in only case of `JIRA` issues. 

Maybe I should change the `JIRA_HOST` to `HOST_URL` and that should be assigned by all the sub-classes. 

Maybe `JIRA_HOST` and `restTemplate`  should be assigned using a constructor in the subclasses?

alecharp: > Having two different methods is important here because the implementation to count issues from both JIRA and GitHub have nothing in common.

Let me just say, what if tomorrow there is a third issue tracker used in the Jenkins project, what would we have to add or changed in here to make it fit? 

In OOP, it should require a minimal addition, very specific. 

If you look at the DependencyBot abstract class, how are the different bot handled?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbe.java
```diff
@@ -0,0 +1,70 @@
...
+@Component
+@Order(value = IssueTrackerDetectionProbe.ORDER)
+class IssueTrackerDetectionProbe extends Probe {
+    public static final String KEY = "issue-tracker-detection";
+    public static final int ORDER = UpdateCenterPluginPublicationProbe.ORDER + 100;
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        context.setIssueTrackType(getIssueTrackerData(context.getUpdateCenter().plugins()));
+        return ProbeResult.success(key(), "Issue tracker detected and returned successfully.");
```
alecharp, line 45: What if the plugin is not in the update-center anymore?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/CodeCoverageProbeTest.java
```diff
@@ -114,7 +114,7 @@ public void shouldBeInErrorWhenRepositoryIsNotInOrganization() {
             Map.of(
                 pluginName, new io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin(
                     pluginName, new VersionNumber("1.0"), scmLink, ZonedDateTime.now(), List.of(), 0,
-                    "42", defaultBranch
+                    "42", defaultBranch, List.of(new io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin.IssueTrackers("", "", ""))
```
alecharp, line 117: You can simply use `List.of()` when you don't care about the existence of the issue trackers.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbeTest.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import io.jenkins.pluginhealth.scoring.model.Plugin;
+import io.jenkins.pluginhealth.scoring.model.ProbeResult;
+import io.jenkins.pluginhealth.scoring.model.updatecenter.UpdateCenter;
+
+import org.junit.jupiter.api.Test;
+
+class IssueTrackerDetectionProbeTest  extends AbstractProbeTest<IssueTrackerDetectionProbe> {
```
alecharp, line 45:
```suggestion
class IssueTrackerDetectionProbeTest extends AbstractProbeTest<IssueTrackerDetectionProbe> {
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbeTest.java
```diff
@@ -0,0 +1,174 @@
...
+        ));
+
+        final IssueTrackerDetectionProbe probe = getSpy();
+
+        assertThat(probe.apply(plugin, ctx))
+            .usingRecursiveComparison()
+            .comparingOnlyFields("id", "status", "message")
+            .isEqualTo(ProbeResult.success(IssueTrackerDetectionProbe.KEY, "Issue tracker detected and returned successfully."));
+
+        assert Objects.equals(ctx.getIssueTrackerType(), Map.of("github", "https://github.com/foo-plugin/issues", "jira", "https://issues.jenkins.io/issues/?jql=component=18331"));
```
alecharp, line 102: Why aren't you assertj function here?

alecharp (changes requested):

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractOpenIssuesProbe.java
aneveux: Hey @Jagrutiti 👋 

Sorry for the late reply, I'm out of the office right now and spend less time on the computer 😝 

Adrien actually gave a nice indication regarding the design: what happens if you had to add a third issue tracker?

The idea behind code abstraction is that the `abstract` class that you write should not know at all about its possible implementations. In your case, not only your `abstract` class knows about Jira and GitHub, but it even provides code for it.

An `abstract` class should contain all the code that is supposed to be common between all of its subclasses, and no more. If there is some code that is used by only one of the subclasses, it is rarely a good fit for the `abstract` class. 

That being said though, maybe while implementing your code you will notice that there is nothing to share between the two classes that you write because they are too different from each other, and in that case, it would be better to challenge the need for abstraction, and simply write 2 classes that are doing their job. 

Introducing abstraction just for the sake of it just makes the source code more complicated, hence less readable.

It is tempting to want to use good practices like abstraction while writing code when we manipulate things that seem similar, but it is important to check that they are indeed similar.

Hope that helps you understand my concern with the class design.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbe.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        Map<String, String>  issueTrackerDetails = getIssueTrackerData(context.getUpdateCenter().plugins());
+
+        if (CollectionUtils.isEmpty(issueTrackerDetails)) {
+            return ProbeResult.failure(key(), String.format("%s plugin does not exists in Update Center.", plugin.getName()));
+        }
+        context.setIssueTrackerNameAndUrl(getIssueTrackerData(context.getUpdateCenter().plugins()));
+        return ProbeResult.success(key(), "Issue tracker detected and returned successfully.");
```
aneveux, line 51: I'm not sure this probe does what you would like it to do.

This probe, for a plugin X, should check if there is an issue tracker that is configured for plugin X. It does not seem that it is doing that from the source code.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbe.java
```diff
@@ -0,0 +1,76 @@
...
+    public String getDescription() {
+        return "Detects the issues tracker type from Update Center.";
+    }
+
+    @Override
+    public String[] getProbeResultRequirement() {
+        return new String[]{ UpdateCenterPluginPublicationProbe.KEY };
+    }
+
+    private Map<String, String> getIssueTrackerData(Map<String, io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin> plugin) {
```
aneveux, line 69: Why is `plugin` a `Map<String, io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin> plugin`? There is a type `Plugin` so a Map should probably be something else than a `plugin`?

Also, why are you using fully qualified names for the `Plugin` type?

aneveux: Also `getIssueTrackerData`, `data` is a bit too vague don't you think? What data exactly are you extracting?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestTemplate;
+
+@Component
+@Order(AbstractOpenIssuesProbe.ORDER)
+class JiraOpenIssuesProbe extends AbstractOpenIssuesProbe {
+    public static final String KEY = "jira-open-issues";
+    private static final String JIRA_HOST = "https://issues.jenkins.io/rest/api/latest/search?";
+    private static final Logger LOGGER = LoggerFactory.getLogger(JiraOpenIssuesProbe.class);
+    RestTemplate restTemplate = new RestTemplate();
```
aneveux, line 48: `RestTemplate` is quite a legacy HTTP client in Spring, nowadays `WebClient` is preferred. 

Since Java 11, there is also a http client that is included in the JDK: `HttpClient`.

Why did you choose `RestTemplate` in comparison with the other preferred alternatives?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * Get total number of open JIRA issues in a plugin
+     */
+    @Override
+    Optional<Integer> getCountOfOpenIssues(ProbeContext context) {
+        String viewJiraIssuesUrl = context.getIssueTrackerNameAndUrl().get("jira");
+
+        try {
+            if (viewJiraIssuesUrl == null || viewJiraIssuesUrl.isEmpty()) {
+                LOGGER.error("JIRA issues not found in Update Center for the plugin");
```
aneveux, line 59: Why is that an error that a plugin is not using Jira?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,94 @@
...
+                return Optional.empty();
+            }
```
aneveux, line 61: This could go outside of the `try` I guess?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,94 @@
...
+            URL url = new URL(viewJiraIssuesUrl);
+            String api = JIRA_HOST.concat(url.getQuery()).concat(" AND status=open");
```
aneveux, line 63: `JIRA_HOST.concat(url.getQuery())` this looked a bit weird to me, isn't the `url` object already containing everything? 

Could be interesting to add in a comment examples of the data you retrieve and why you arrange them this way.

aneveux: Also, is it OK to have spaces in ` AND status=open`? Are those spaces going to be properly escaped?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,94 @@
...
+
+            ResponseEntity<String> response = restTemplate.getForEntity(api, String.class);
+            ObjectMapper objectMapper = new ObjectMapper();
```
aneveux, line 66: This will create one `ObjectMapper` per probe, while you could use only one instance that you share, as written in the Javadocs: https://javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/2.9.8/com/fasterxml/jackson/databind/ObjectMapper.html

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,94 @@
...
+            String jsonResponse = response.getBody();
+            JsonNode jsonNode = objectMapper.readTree(jsonResponse);
+
+            if (jsonNode.get("errorMessages") != null) {
+                LOGGER.error("Error returned from JIRA API for plugin {}", jsonNode.get("errorMessages"));
+                return Optional.empty();
+            }
+            return Optional.of(jsonNode.get("total").asInt());
+        } catch (JsonMappingException e) {
+            LOGGER.error("Cannot map JSON returned by JIRA API for plugin ", e);
```
aneveux, line 76: You forgot the `{}` at the end of the message?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
```diff
@@ -88,6 +89,14 @@ public Optional<String> getRepositoryName(String scm) {
         return match.find() ? Optional.of(match.group("repo")) : Optional.empty();
     }
 
+    public void setIssueTrackerNameAndUrl(Map<String, String> issueTrackerNameAndUrl) {
+        this.issueTrackerNameAndUrl = issueTrackerNameAndUrl;
+    }
+
+    public Map<String, String> getIssueTrackerNameAndUrl() {
```
aneveux, line 96: Actually with `getIssueTrackerNameAndUrl` as a method name I would expect either to receive a `String` or just a `Tuple` or a custom object, but not a Map, so I guess using `issueTrackers` as a name would make more sense?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/CodeCoverageProbeTest.java
```diff
@@ -114,7 +114,7 @@ public void shouldBeInErrorWhenRepositoryIsNotInOrganization() {
             Map.of(
                 pluginName, new io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin(
                     pluginName, new VersionNumber("1.0"), scmLink, ZonedDateTime.now(), List.of(), 0,
-                    "42", defaultBranch
+                    "42", defaultBranch, List.of()
```
aneveux, line 117: Looks like you had to modify a lot of similar data? Maybe it's worth extracting somewhere so you would only have 1 place to update?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/GitHubOpenIssuesProbeTest.java
```diff
@@ -0,0 +1,136 @@
...
+    void shouldBeAbleToFindNumberOfOpenIssuesInGH() throws IOException {
+        final String pluginName = "cloudevents";
+        final String repository = "jenkinsci/" + pluginName + "-plugin";
+        final String scmLink = "https://github.com/" + repository;
+
+        final Plugin plugin = mock(Plugin.class);
+        final ProbeContext ctx = mock(ProbeContext.class);
+        final GitHub gh = mock(GitHub.class);
+        final GHRepository ghRepository = mock(GHRepository.class);
+        final io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin.IssueTrackers issueTrackerGithub = new io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin.IssueTrackers("github", "https://github.com/" + repository + "/issues", "https://github.com/" + repository + "/issues/new/choose");
```
aneveux, line 98: Why not importing the type and using a fully qualified name here?

aneveux, line 98: Why are you concatenating again `"https://github.com/" + repository + "/issues"` when you already have the first part in `scmLink`?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractOpenIssuesProbe.java
Jagrutiti: Thank you for explaining @aneveux.

In past cases when I used `abstract` classes. 99% of the code was common. 

In this case, there was no common code. Maybe I should have asked this in the meeting itself but it did not occur to me then.

I should have taken inspiration from `doApply` abstract method in `Probe` class. But I did not understand the "why" part until I made this blunder.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbe.java
```diff
@@ -0,0 +1,76 @@
...
+
+@Component
+@Order(value = IssueTrackerDetectionProbe.ORDER)
+class IssueTrackerDetectionProbe extends Probe {
+    public static final String KEY = "issue-tracker-detection";
+    public static final int ORDER = UpdateCenterPluginPublicationProbe.ORDER + 100;
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        Map<String, String>  issueTrackerDetails = getIssueTrackerData(context.getUpdateCenter().plugins());
```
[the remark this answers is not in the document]
Jagrutiti: I usually use `CTRL + ALT + L` in IntelliJ.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractOpenIssuesProbe.java
aneveux: That's alright @Jagrutiti =) We learn by making mistakes, and if you never experienced such a case before, it would have been difficult to think about it and ask a question during the meeting.

If you understand it better now, then that's great =)

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
Jagrutiti: It could be an `info`?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
Jagrutiti: Yes, they are escaped in the browser.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
Jagrutiti: I am curious, how do you expect to receive a particular datatype as a parameter based on getter and setter method name? You have said similar things in previous reviews as well. 

In the meeting, Adrien said that the name wasn't descriptive enough. And when he asked me what the method returns, I forgot too. Until I re-read my test cases. That is why I renamed them this way to remember that it returns `type` and `viewUrl` from IssueTrackers. 

The `ProbeContext` class has no JavaDocs for any of its members. Shall I add one in this case? Then I guess renaming won't be a problem.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DeprecatedPluginProbeTest.java
```diff
@@ -62,7 +62,7 @@ void shouldBeAbleToDetectNonDeprecatedPlugin() {
 
         when(plugin.getName()).thenReturn("foo");
         when(ctx.getUpdateCenter()).thenReturn(new UpdateCenter(
-            Map.of("foo", new Plugin("foo", new VersionNumber("1.0"), "scm", ZonedDateTime.now().minusDays(1), Collections.emptyList(), 0, "", "main")),
+            Map.of("foo", new Plugin("foo", new VersionNumber("1.0"), "scm", ZonedDateTime.now().minusDays(1), Collections.emptyList(), 0, "", "main", List.of())),
```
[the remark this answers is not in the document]
Jagrutiti: I tried to understand constructors in Java records. When I came across this article: https://blogs.oracle.com/javamagazine/post/records-come-to-java#anchor_8

In this article, the author speaks of alternative constructors and says I can return default data something like this.

```
    public static Plugin of(String name, VersionNumber version, String scm,
                            ZonedDateTime releaseTimestamp, List<String> labels,
                            int popularity, String requiredCore, String defaultBranch) {
        return new Plugin(name, version, scm, releaseTimestamp,
            labels, popularity, requiredCore, defaultBranch, List.of());
    }    
```
   
    
   And in the test case, I should call it this  way:
   
```
 when(ctx.getUpdateCenter()).thenReturn(new UpdateCenter(
            Map.of("foo", Plugin.of("foo", new VersionNumber("1.0"), "scm", ZonedDateTime.now().minusDays(1), Collections.emptyList(), 0, "", "main")),
            Map.of("bar", new Deprecation("find-the-reason-here")),
            Collections.emptyList()
        ));
```
   
  I will still need to remove the `new` keyword and add `.of` to `Plugin.`

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/GitHubOpenIssuesProbeTest.java
Jagrutiti: For some reason, IntelliJ did not add a qualified import. Maybe because of the common `Plugin` class.

In the case of the common `Plugin` class name in UpdateCenter and Model, it did that. That is why I did not think much of it.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
Jagrutiti: I googled "calling api in spring boot" most of them example of `RestTemplate`. That is why I used this.

Jagrutiti:
Thank you for your detailed review @aneveux.

It was overwhelming at first. But I considered each point and fixed the shortcomings. 

Thank you for your time especially as you are on vacation.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,128 @@
...
+        try {
+            // The `url` will contain the JIRA url to view issues. For ex: https://issues.jenkins.io/rest/api/latest/search?jql=component=15979
+            URL url = new URL(viewJiraIssuesUrl);
+
+            /* Here, the query of the url "?jql=component=1833" is concatenated with " AND status=open".
+               This gives the final API required to fetch JIRA issues.
+               For ex: https://issues.jenkins.io/rest/api/latest/search?jql=component=15979%20and%20status=open
+             */
+            String api = JIRA_HOST.concat(url.getQuery())
+                        .concat(URLEncoder.encode(" AND ", StandardCharsets.UTF_8.toString()).replace("+", "%20")
```
Jagrutiti, line 85: Is it okay if I replace this line `.concat(URLEncoder.encode(" AND ", StandardCharsets.UTF_8.toString()).replace("+", "%20")` with just `%20AND%20`?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
aneveux: That is not very surprising cause `RestTemplate` is there for quite some time.

I would advise to switch to a more up-to-date alternative though, in order to avoid introducing technical debt in new code additions.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
aneveux: What I meant is that it should not be considered an issue at all right? Cause a plugin could be using GitHub issues instead of Jira, and that should neither be a warning nor an error?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ProbeContext.java
aneveux: You are right, this sounds like preferences.

There are actually bunch of articles talking about good practices linked to naming in Java (or in languages in general). Considering code is much more read than written, it is important to optimize it for reading (code should be understandable easily and maintainable) rather than optimizing its writing.

In this particular context, what triggered me is the following: when I read `IssueTrackerNameAndUrl` I expect the method to return ONLY the issue tracker name and its url, but here the type that is returned is a `Map` which could contain more things, and which can actually contain several issue trackers. 

Naming is a very hard part of writing code, and it is quite a subjective thing, so it is normal to have discussions about that. IMHO, there are some global rules you can follow: plurals usually means you are talking about collections, etc.

For this particular variable, IMHO we could either go for something like `issueTrackers` but you are right that it might not be descriptive enough, so in that case I would usually go for something like `issueTrackerUrlsByNames` where I describe the content of the Map and how it is indexed.

Javadocs are also a good way to describe what is in variables, and give extra details, but that does not replace good naming (we should aim for both :stuck_out_tongue:)

aneveux:
> It was overwhelming at first. But I considered each point and fixed the shortcomings.

Sorry about that :/ 

Just to be clear, I do not want to undermine your efforts, I appreciate your reactions to reviews, and I am sharing lots of comments wanting to help you discovering and learning things, but if you'd like me to do it some other way, just let me know.

Your approach of taking the comments one by one is good though! Keep up the great work :muscle:

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
aneveux: Sure =)

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
Jagrutiti: I will make it an `info`?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,127 @@
...
+               This gives the final API required to fetch JIRA issues.
+               For ex: https://issues.jenkins.io/rest/api/latest/search?jql=component=15979%20and%20status=open
+             */
+            String api = JIRA_HOST.concat(url.getQuery())
+                .concat("%20AND%20")
+                .concat("status=open");
+
+            httpRequest = HttpRequest.newBuilder()
+                .uri(new URI(api))
+                .timeout(Duration.of(5, SECONDS))
```
aneveux, line 89: Isn't this too quick? What is the default value?

IMHO, for any decision like this, it is interesting that you leave a simple comment in the source code explaining why you took that decision, so readers can understand if it is a deliberate choice, and what is the rationale for it.

General remark: remember you mostly write code for other humans, so you need to tell them your intention.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/CodeCoverageProbeTest.java
```diff
@@ -112,7 +112,7 @@ public void shouldBeInErrorWhenRepositoryIsNotInOrganization() {
 
         when(ctx.getUpdateCenter()).thenReturn(new UpdateCenter(
             Map.of(
-                pluginName, new io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin(
+                pluginName, io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin.of(
```
aneveux, line 115: Looks like a lot of those `Plugin` objects are the same, with similar values. Maybe extract that in a variable that can be reused everywhere?

aneveux (commented):
Few comments from a quick review. General remark: adding Javadocs on the classes you write is also interesting to understand the rationale for classes to exist. It is often a nice thing to read as a first discovery of what a class is for.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
Jagrutiti: The [JavaDoc ](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpRequest.Builder.html#timeout(java.time.Duration))says:

>  The effect of not setting a timeout is the same as setting an infinite Duration, i.e. block forever.

I was not really sure which value to pick. Honestly, the APIs I had tested took less than a second to respond. So I figured 5 seconds is apt.

What should I actually write as the comment?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/CodeCoverageProbeTest.java
Jagrutiti: The [JavaDoc ](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpRequest.Builder.html#timeout(java.time.Duration))says 

>  The effect of not setting a timeout is the same as setting an infinite Duration, i.e. block forever.

I was not really sure which value to pick. Honestly, the APIs I had tested took less than a second to respond. So I figured 5 seconds is apt.

What should I actually write as the comment?

Jagrutiti: Here  parameters like `pluginName`, `scmLink`, `defaultBranch` are mocked before being passed. 

You do not mean creating a constructor in Record like I already did it once?

core/pom.xml
```diff
@@ -102,5 +102,11 @@
 			<artifactId>postgresql</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<!-- Used to invoke API calls  -->
+		<dependency>
+			<groupId>org.apache.httpcomponents</groupId>
+			<artifactId>httpclient</artifactId>
+			<version>4.5.14</version>
+		</dependency>
```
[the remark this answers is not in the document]
Jagrutiti: Apache HTTPClient is an additional dependency. While I can do the same thing using Java HTTPClient. So I shouldn't be using it.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
aneveux: > What should I actually write as the comment?

The general rule to follow while writing source code is that someone reading it should not have to guess why you did certain things, but should find the information written explicitely.

If you thought that 5 seconds was a good timeout value based on your testing, then write it in a simple comment, so it is now our shared understanding.

Basically, imagine reading that piece of source code if you did not write it yourself, and try to see if you understand everything from what you read.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbe.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        Map<String, String> issueTrackerDetails = getIssueTrackersForAPlugin(plugin.getName(), context.getUpdateCenter());
+
+        if (issueTrackerDetails.isEmpty()) {
+            return ProbeResult.failure(key(), String.format("No issue tracker data available for %s plugin in Update Center.", plugin.getName()));
+        }
+        context.setIssueTrackerNameAndUrl(issueTrackerDetails);
+        return ProbeResult.success(key(), "Issue tracker detected and returned successfully.");
```
aneveux, line 53: I am not entirely convinced about the message here. IIRC, it is used in the GUI to display the status to users, and I feel it is a bit too technical (returned successfully).

Could be something like `Found %d issue trackers configured for the plugin.` (cause I believe we usually not repeat the plugin name in the message?)

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbe.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Gets issue trackers for a specific plugin from Update Center.
+     *
+     * @param pluginName   name of the plugin to fetch issue tracker data for.
+     * @param updateCenter The {@link UpdateCenter}.
+     * @return A Map of filtered data from issue trackers.
+     */
+    private Map<String, String> getIssueTrackersForAPlugin(String pluginName, UpdateCenter updateCenter) {
+        return updateCenter.plugins().get(pluginName) != null
+            ? filterIssueTrackersForTypeAndViewUrl(updateCenter.plugins().get(pluginName).getIssueTrackers())
```
aneveux, line 80: Can this: `.getIssueTrackers()` be `null`?

If so, you'll have a NPE line 91.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,127 @@
...
+
+@Component
+@Order(AbstractOpenIssuesProbe.ORDER)
+class JiraOpenIssuesProbe extends AbstractOpenIssuesProbe {
+    public static final String KEY = "jira-open-issues";
+    private static final String JIRA_HOST = "https://issues.jenkins.io/rest/api/latest/search?";
+    private static final Logger LOGGER = LoggerFactory.getLogger(JiraOpenIssuesProbe.class);
+    final ObjectMapper objectMapper = new ObjectMapper();
+    HttpClient httpClient;
+    HttpRequest httpRequest;
```
aneveux, line 57: I feel those should go in the `getCountOfOpenIssues` method and not in the class itself. They are not meant to be reused by something else in the class right?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Get total number of open JIRA issues in a plugin.
+     *
+     * @param context {@link ProbeContext}
+     * @return Optional of type Integer.
+     */
+    @Override
+    Optional<Integer> getCountOfOpenIssues(ProbeContext context) {
+        String viewJiraIssuesUrl = context.getIssueTrackerUrlsByNames().get("jira");
```
aneveux, line 67: `context.getIssueTrackerUrlsByNames()` cannot be `null` right?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,127 @@
...
+
+        if (viewJiraIssuesUrl == null || viewJiraIssuesUrl.isEmpty()) {
+            LOGGER.info("The plugin does not use JIRA to track issues.");
```
aneveux, line 70: Same remark: which plugin?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,127 @@
...
+                .timeout(Duration.of(5, SECONDS)) // based on manual testing, timeout after 5 seconds works.
+                .GET()
+                .build();
+
+            HttpResponse<String> response = httpClient.send(httpRequest, HttpResponse.BodyHandlers.ofString());
+            String jsonResponse = response.body();
+            JsonNode jsonNode = objectMapper.readTree(jsonResponse);
+
+            if (jsonNode.get("errorMessages") != null) {
+                LOGGER.error("Cannot request JIRA API for the plugin. {}", jsonNode.get("errorMessages"));
```
aneveux, line 98: Which plugin?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,127 @@
...
+            LOGGER.error("Cannot process JSON returned by JIRA API for the plugin. {}", e);
+        } catch (MalformedURLException e) {
+            LOGGER.error("Cannot process malformed URL for the plugin. {}", e);
+        } catch (URISyntaxException e) {
+            LOGGER.error("Incorrect URI syntax in the plugin. {}", e);
+        } catch (IOException e) {
+            LOGGER.error("Cannot read HttpResponse for the plugin. {}", e);
+        } catch (InterruptedException e) {
+            LOGGER.error("Interruption occurred when waiting for JIRA API for the plugin. {}", e);
+        }
```
aneveux, line 114: Which plugin?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/CodeCoverageProbeTest.java
aneveux: That's OK like this

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/GitHubOpenIssuesProbeTest.java
```diff
@@ -0,0 +1,182 @@
...
+        when(ctx.getUpdateCenter()).thenReturn(new UpdateCenter(
+            Map.of(pluginName, new io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin(
+                pluginName, null, null, null, List.of(), 0, "2.361.1", "main",
+                List.of(issueTrackerGithub)
+            )),
+            Map.of(),
+            List.of()
+        ));
+
+        when(ctx.getIssueTrackerUrlsByNames()).thenReturn(Map.of("github", "https://github.com/" + repository + "/issues"));
```
aneveux, line 119:
```suggestion
        when(ctx.getIssueTrackerUrlsByNames()).thenReturn(Map.of("github", scmLink + "/issues"));
```

aneveux (commented):
Very minor comments. Great job Jagruti!

Just few little things to clean-up and we'll be ready to merge.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/GitHubOpenIssuesProbe.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Get total number of open GitHub issues in a plugin.
+     *
+     * @param context {@link ProbeContext}
+     * @return Optional an Integer value will give total count of open issues.
+     */
+    @Override
+    Optional<Integer> getCountOfOpenIssues(ProbeContext context) {
+        /* Stores the GitHub URL to view all existing issues in the plugin. Ex: https://github.com/jenkinsci/cloudevents-plugin/issues */
+        String issueTrackerViewUrl = context.getIssueTrackerUrlsByNames().get("github");
```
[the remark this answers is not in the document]
Jagrutiti: `IssueTrackerDetectionProbe` is the requirement for `GitHubOpenIssuesProbe` and `UpdateCenterPluginPublicationProbe` is the requirement for `IssueTrackerDetectionProbe` that is why a part of my believes that this issue might not occur. 

It is better to be safe than sorry.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/GitHubOpenIssuesProbe.java
```diff
@@ -0,0 +1,79 @@
...
+
+        if (issueTrackerViewUrl == null) {
+            LOGGER.info("The plugin does not use GitHub issues to track issues.");
```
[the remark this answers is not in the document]
Jagrutiti: Hi @aneveux 

About adding the names of the plugin, in the logger message, Adrien told me that I shouldn't be passing a parameter to the method just to add it in the logger message. (I had made the mistake in this probe itself and I am not proud of it)

`ProbeContext` does not contain the plugin details like names. I would need to pass the `Plugin` parameter here which is required only in the logs. I do not think this is the best approach. 

After the logs are printed and when the `ProbeResult` is returned as the final result, I have added the plugin name in both the cases. Failure as well as the success message. 

So when the user reads the logs and sees the final `ProbeResult` message I am sure it would make sense to them. 

I hope I understood your concern correctly. If I am on the wrong track here do let me know :)

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbe.java
Jagrutiti: In this case `IssueTracker` becomes the name of the plugin and also "issue tracker" is what the whole exercise of fetching data from the `IssueTrackers`. That is why it sounds a little ambiguous.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/GitHubOpenIssuesProbe.java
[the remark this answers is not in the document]
aneveux: Is there a log message with the plugin name at the beginning of the process?

Something like "Running probes for plugin XXX" and then you have all the logs of the probes execution for that XXX plugin?

I do understand the concern of having the Plugin as parameter just for logging, indeed not a good thing to do.

Anyway, that's probably nitpicking then, so let's not focus too much on that.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/GitHubOpenIssuesProbe.java
```diff
@@ -47,6 +47,10 @@ class GitHubOpenIssuesProbe extends AbstractOpenIssuesProbe {
      */
     @Override
     Optional<Integer> getCountOfOpenIssues(ProbeContext context) {
+        if (context.getIssueTrackerUrlsByNames().get("github") == null) {
```
aneveux, line 50: What I meant was to check that:

`if (context.getIssueTrackerUrlsByNames() == null)`

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -64,6 +64,12 @@ class JiraOpenIssuesProbe extends AbstractOpenIssuesProbe {
      */
     @Override
     Optional<Integer> getCountOfOpenIssues(ProbeContext context) {
+        if (context.getIssueTrackerUrlsByNames().get("jira") == null) {
```
aneveux, line 67: Same here, I meant `if (context.getIssueTrackerUrlsByNames() == null) {`

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/GitHubOpenIssuesProbe.java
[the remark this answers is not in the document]
Jagrutiti: > Is there a log message with the plugin name at the beginning of the process?

Not really. In the project I have not seen info log at the beginning of the probe. I am not sure if I should add this.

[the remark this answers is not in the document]
aneveux: Let's not take care of that in that Pull Request then, could be something to have a look at later on while harmonizing the log messages.

aneveux (approved):
LGTM.

I'll let @alecharp have another look at it in case I missed something, and I believe that'll be ready to merge :rocket:

core/src/main/java/io/jenkins/pluginhealth/scoring/model/updatecenter/Plugin.java
```diff
@@ -31,8 +31,31 @@
...
+    }
+
     public io.jenkins.pluginhealth.scoring.model.Plugin toPlugin() {
         return new io.jenkins.pluginhealth.scoring.model.Plugin(this.name(), this.version(), this.scm(), this.releaseTimestamp());
     }
+
+    public List<IssueTrackers> getIssueTrackers() {
+        return this.issueTrackers;
+    }
+
```
alecharp, line 51: You don't need this on records. "Getters" are created by default. The name of the method is `issueTrackers()` like the field specified in the record definition. See https://docs.oracle.com/en/java/javase/17/language/records.html

```suggestion

```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/GitHubOpenIssuesProbe.java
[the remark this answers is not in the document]
alecharp: I like the "better safe than sorry" principal. 👍

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IssueTrackerDetectionProbe.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Gets issue trackers for a specific plugin from Update Center.
+     *
+     * @param pluginName   name of the plugin to fetch issue tracker data for.
+     * @param updateCenter The {@link UpdateCenter}.
+     * @return A Map of filtered data from issue trackers.
+     */
+    private Map<String, String> getIssueTrackersForAPlugin(String pluginName, UpdateCenter updateCenter) {
+        return (updateCenter.plugins().get(pluginName) != null) && (updateCenter.plugins().get(pluginName).getIssueTrackers() != null)
+            ? filterIssueTrackersForTypeAndViewUrl(updateCenter.plugins().get(pluginName).getIssueTrackers())
```
alecharp, line 80:
```suggestion
            ? filterIssueTrackersForTypeAndViewUrl(updateCenter.plugins().get(pluginName).issueTrackers())
```

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
alecharp: I agree here. those fields should be local variables.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
alecharp: Same as before, "better safe than sorry".

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
alecharp: I don't think this was addressed.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
```diff
@@ -0,0 +1,133 @@
...
+
+        if (viewJiraIssuesUrl == null || viewJiraIssuesUrl.isEmpty()) {
+            LOGGER.info("The plugin does not use JIRA to track issues.");
+            return Optional.empty();
+        }
+        try {
+            /* The `url` will contain the JIRA url to view issues.
+               For ex: https://issues.jenkins.io/rest/api/latest/search?jql=component=15979
+            */
+            URL url = new URL(viewJiraIssuesUrl);
```
alecharp, line 83: Because `URL#equals` method is doing a DNS resolution, I'd prefer using `URI` class which provide the same features but with less problems.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/CodeCoverageProbeTest.java
```diff
@@ -49,16 +49,16 @@
...
     @Test
     public void shouldNotRequireRelease() {
         assertThat(getSpy().requiresRelease()).isFalse();
     }
 
+    @Override
+    CodeCoverageProbe getSpy() {
+        return spy(CodeCoverageProbe.class);
+    }
+
```
alecharp, line 61: Why moving methods around?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DeprecatedPluginProbeTest.java
```diff
@@ -44,16 +44,16 @@
...
     @Test
     void shouldNotRequireRelease() {
         assertThat(getSpy().requiresRelease()).isFalse();
     }
 
+    @Override
+    DeprecatedPluginProbe getSpy() {
+        return spy(DeprecatedPluginProbe.class);
+    }
+
```
alecharp, line 56: Same.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JenkinsCoreProbeTest.java
```diff
@@ -42,16 +42,16 @@
...
     @Test
     void shouldRequireRelease() {
         assertThat(getSpy().requiresRelease()).isTrue();
     }
 
+    @Override
+    JenkinsCoreProbe getSpy() {
+        return spy(JenkinsCoreProbe.class);
+    }
+
```
alecharp, line 54: Same.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbeTest.java
```diff
@@ -0,0 +1,196 @@
...
+import io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin.IssueTrackers;
+import io.jenkins.pluginhealth.scoring.model.updatecenter.UpdateCenter;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.InjectMocks;
+
+class JiraOpenIssuesProbeTest extends AbstractProbeTest<JiraOpenIssuesProbe> {
+
+    @InjectMocks
+    JiraOpenIssuesProbe jiraOpenIssuesProbe = new JiraOpenIssuesProbe();
```
alecharp, line 53: Why? This is not supposed to work as you didn't add the annocation `@ExtendWith(MockitoExtension.class)` on the test class, so this seems to be something you added to try to make the test work, but I don't see it used anywhere.

```suggestion
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbeTest.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.List;
+import java.util.Map;
+
+import io.jenkins.pluginhealth.scoring.model.Plugin;
+import io.jenkins.pluginhealth.scoring.model.ProbeResult;
+import io.jenkins.pluginhealth.scoring.model.updatecenter.Plugin.IssueTrackers;
+import io.jenkins.pluginhealth.scoring.model.updatecenter.UpdateCenter;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.InjectMocks;
```
alecharp, line 48:
```suggestion
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/KnownSecurityVulnerabilityProbeTest.java
```diff
@@ -45,16 +45,16 @@
...
     @Test
     void shouldNotRequireNewRelease() {
         assertThat(getSpy().requiresRelease()).isFalse();
     }
 
+    @Override
+    KnownSecurityVulnerabilityProbe getSpy() {
+        return spy(KnownSecurityVulnerabilityProbe.class);
+    }
+
```
alecharp, line 57: Same as before.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SpotBugsProbeTest.java
```diff
@@ -50,16 +50,16 @@
...
     @Test
     public void shouldNotRequireRelease() {
         assertThat(getSpy().requiresRelease()).isFalse();
     }
 
+    @Override
+    protected SpotBugsProbe getSpy() {
+        return spy(SpotBugsProbe.class);
+    }
+
```
alecharp, line 62: Again.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/UpForAdoptionProbeTest.java
```diff
@@ -43,16 +43,16 @@
...
     @Test
     void shouldNotRequireNewRelease() {
         assertThat(getSpy().requiresRelease()).isFalse();
     }
 
+    @Override
+    UpForAdoptionProbe getSpy() {
+        return spy(UpForAdoptionProbe.class);
+    }
+
```
alecharp, line 55: Again.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/UpdateCenterPluginPublicationProbeTest.java
```diff
@@ -40,16 +40,16 @@
...
     @Test
     void shouldNotRequireRelease() {
         assertThat(getSpy().requiresRelease()).isFalse();
     }
 
+    @Override
+    UpdateCenterPluginPublicationProbe getSpy() {
+        return spy(UpdateCenterPluginPublicationProbe.class);
+    }
+
```
alecharp, line 52: Again.

alecharp (changes requested):
there are a few things that were not addressed.

I would like to know why you moved the `getSpy` method implementation on most (if not all) test classes.

Jagrutiti:
> there are a few things that were not addressed.
> 
> I would like to know why you moved the `getSpy` method implementation on most (if not all) test classes.

All I did was `Ctrl + Alt + L`. It moved because of formatting.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
Jagrutiti: I cannot mock the variables if they are not global in the test case.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbeTest.java
Jagrutiti: I am using it for statements like this one `jiraOpenIssuesProbe.httpClient = mockedHttpClient;`

Jagrutiti: You are right the `@InjectMocks` annotation's availability does not make a difference to the tests.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/SpotBugsProbeTest.java
Jagrutiti: It is all because of `Ctrl + Alt + L`. I did not do it manually.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/CodeCoverageProbeTest.java
Jagrutiti: It is automated formatting. That is why I did not change it.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
Jagrutiti: This requires the `plugin` to be passed as a parameter to `getCountOfOpenIssues` to get `plugin.getName()` that is why 
 did not  work on this one.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
Jagrutiti: The `if` condition checks whether the `null` values below this line.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
alecharp: or you could just provide the `name` of the plugin, rather than the entire `plugin`.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
alecharp: ok. Let me look into this because this doesn't seem right to create class fields just because of the test.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JiraOpenIssuesProbe.java
Jagrutiti: > or you could just provide the `name` of the plugin, rather than the entire `plugin`.

I did something similar last time. You said we shouldn't pass a parameter just to display in the logs, But if you are cool with it I  can do that,

alecharp:
Is this still something you want to contribute @Jagrutiti?

Jagrutiti:
> Is this still something you want to contribute @Jagrutiti?

Yes. I am interested. 

I remember this PR wasn't merged because of some code structuring thing and variable scope used in the test cases that you wanted to refactor.

alecharp:
@Jagrutiti do you think you will be able to complete this pull request?

alecharp:
Closing this for now.

PR #358, opened 2023-08-20
Incremental Build detection probe

### LINKED ISSUE #321: Incremental build detection
### Description

It is possible for Jenkins plugin to have one commit been published as a release. This is part of [JEP-305](https://github.com/jenkinsci/jep/blob/master/jep/305/README.adoc).

Jagrutiti:
### Description

This probe detects whether incremental build is configured in the plugin.

Closes #321 

### Testing done

```[tasklist]
### Submitter checklist
```


Jagrutiti:
My rationale behind designing two probes for Incremental Build:

In the Gitter channel, @aneveux suggested working on detecting constants that do not change:

- > An extensions.xml file in .mvn that contains the io.jenkins.tools.incrementals extension, like [this one in the kubernetes-plugin](https://github.com/jenkinsci/kubernetes-plugin/blob/0141a5f145128b5b8458d40686a3904006f45b8d/.mvn/extensions.xml) for example.
- > A maven.config file in .mvn containing the consume-incrementals and might-produce-incrementals profiles (it could contain more things, we don't care) like [this one in kubernetes-plugin](https://github.com/jenkinsci/kubernetes-plugin/blob/0141a5f145128b5b8458d40686a3904006f45b8d/.mvn/maven.config) for example.
- > A GitHub action that is referring to jenkins-infra/github-reusable-workflows/.github/workflows/maven-cd.yml. We could assume this to be in a cd.yaml file in .github/workflows, but I would rather search for any action that refers to jenkins-infra/github-reusable-workflows/.github/workflows/maven-cd.yml instead to avoid failing a plugin that simply renamed the workflow file, which is totally legit.

To search for `maven-cd.yml` I implemented `AbstractGitHubWorkflowProbe` because it is required in GitHub Actions. It made no sense to duplicate the already existing code.

To detect builds in Maven configuration I went through some of the old probes I have worked on and also the existing probes. 
And decided that in this case designing a second probe is the best option from the perspective of code readability and maintenance. 

For scoring maybe we can do something like check if any one of the probes has a non-zero score. If yes, we do not execute the second probe.

alecharp (changes requested):

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionInGitHubWorkflowProbe.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package io.jenkins.pluginhealth.scoring.probes;
+
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(AbstractGitHubWorkflowProbe.ORDER)
+public class IncrementalBuildDetectionInGitHubWorkflowProbe extends AbstractGitHubWorkflowProbe {
```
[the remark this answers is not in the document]
aneveux: Oh that is correct! I was not sure if CD was considered as part of the incrementals setup or not, now I know :p 

Sorry about that, I might have be misleading a bit on that one.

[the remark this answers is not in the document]
alecharp: Well, this probe was looking for the same workflow definition than cd so...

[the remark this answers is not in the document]
aneveux: (I forgot we had a CD probe :stuck_out_tongue:)

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,134 @@
...
+            LOGGER.error("Could not read the file during probe {}. {}", key(), e);
+        } catch (ParserConfigurationException | SAXException e) {
+            LOGGER.error("Could not parse the file during probe {}. {}", key(), e);
+        }
+        return false;
+    }
+
+    private boolean isMavenConfigConfigured(Path path) {
+        try {
+            return Files.lines(path).anyMatch(line -> line.contains("-Pconsume-incrementals") || line.contains("-Pmight-produce-incrementals"));
```
aneveux, line 128: This is not doing what we want. We want to check that both profiles are present in the file, not only one of them.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbeTest.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Override
+    IncrementalBuildDetectionProbe getSpy() {
+        return spy(IncrementalBuildDetectionProbe.class);
+    }
+
+    @Test
+    void shouldReturnASuccessfulCheckIncrementalBuildConfiguredInBothFiles() {
+        when(ctx.getScmRepository()).thenReturn(Path.of("src/test/resources/jenkinsci/test-plugin-1"));
```
aneveux, line 41: For better readability, please name your test resources in regards to the expected test result.

Something like `properlyConfiguredPlugin`, `pluginMissingExtensions`, `pluginMissingMavenConfig`, `pluginWithMalformedExtension`, etc.

This way it is much easier to read the test.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbeTest.java
```diff
@@ -0,0 +1,98 @@
...
+        assertThat(probe.apply(plugin, ctx))
+            .usingRecursiveComparison()
+            .comparingOnlyFields("id", "message", "status")
+            .isEqualTo(ProbeResult.success(IncrementalBuildDetectionProbe.KEY, "Incremental Build is configured in the plugin."));
+        verify(probe).doApply(plugin, ctx);
+    }
+
+    @Test
+    void shouldReturnASuccessfulCheckWhenIncrementalBuildIsConfiguredOnlyInExtensionsXML() {
```
aneveux, line 50: Not the expected behavior. We need both files to be present, not just one of them.

aneveux (changes requested):
I left a few comments. The start is OK, but you did not implement what we initially discussed about.

While writing your test cases, be careful to check that the test you write indeed tests the expect business scenario.

You are not very far from having this probe properly working, but it requires fixing.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,134 @@
...
+    public static final String KEY = "incremental-build-maven-configuration";
+    private static final Logger LOGGER = LoggerFactory.getLogger(IncrementalBuildDetectionProbe.class);
+    private static final String INCREMENTAL_TOOL = "io.jenkins.tools.incrementals";
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        final Path scmRepository = context.getScmRepository();
+        final Path mvnConfig = scmRepository.resolve(".mvn");
+        if (Files.notExists(mvnConfig)) {
+            LOGGER.error("No Maven configuration folder at {}.", key());
```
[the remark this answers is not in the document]
Jagrutiti: > Remember, an error log should mean that there is something wrong going with the application and that it failed to recover from it.

I need to change my mindset here. I believe  that anything that prevents the the functionality from completing should be an error. I will be more careful about it.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
Jagrutiti: Here changing `||` to `&&` wasn't working. I had to change more of the syntax.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,153 @@
...
+        final Path scmRepository = context.getScmRepository();
+        final Path mvnConfig = scmRepository.resolve(".mvn");
+        if (Files.notExists(mvnConfig)) {
+            LOGGER.info("Could not find Maven configuration folder {} plugin while running {} probe.", plugin.getName(), key());
+            return ProbeResult.failure(key(), String.format("Could not find Maven configuration folder for the %s plugin.", plugin.getName()));
+        }
+
+        try (Stream<Path> paths = Files.find(mvnConfig, 1, (path, $) -> Files.isRegularFile(path))) {
+            Map<String, Path> map = paths
+                .collect(Collectors.toMap(
```
Jagrutiti, line 72: In this case, I was not able to decide on the correct approach. 

Another alternative is using `forEach` or  `map` in that case I can't use `Collectors.toMap` and need an external map that will add the data with the desired key name.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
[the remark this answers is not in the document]
aneveux: https://dev.to/raysaltrelli/logging-best-practices-obo

This is an interesting first reading regarding log levels.

The important thing to keep in mind is that logs should relate to your application's execution, not your application's business implementation.

If a probe is detecting something wrong, it is not an error from a log level, because the probe did run perfectly fine, and accomplished its job of identifying a business error.

If a probe cannot detect something because it encountered an Exception though, and you cannot figure if the project you are analyzing is correct or not, that is an error to log, because your application cannot do what it is supposed to do.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,153 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.xml.sax.SAXException;
+
+/**
+ * This probe checks for Incremental Build configuration in plugins
```
aneveux, line 50: In such Javadocs, it is actually interesting to share a bit more information about the decisions that you took for writing that class, like for examples how you intend to detect that a plugin is using incrementals.

Not really like writing specifications as Javadocs, but rather sharing a bit of the rationale for that class to exist.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,153 @@
...
+ */
+@Component
+@Order(IncrementalBuildDetectionProbe.ORDER)
+public class IncrementalBuildDetectionProbe extends Probe {
+    public static final int ORDER = SCMLinkValidationProbe.ORDER + 100;
+    public static final String KEY = "incremental-build-maven-configuration";
+    private static final Logger LOGGER = LoggerFactory.getLogger(IncrementalBuildDetectionProbe.class);
+    private static final String INCREMENTAL_TOOL = "io.jenkins.tools.incrementals";
+    private static final String ARTIFACT_ID = "git-changelist-maven-extension";
```
aneveux, line 59: `ARTIFACT_ID` is probably a too vague name. What is it the `ArtifactId` of?

Something like `INCREMENTAL_TOOL_ARTIFACT_ID` would be more accurate I believe.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,153 @@
...
+                        if (filePath.endsWith("maven.config")) {
+                            return "maven";
+                        }
+                        if (filePath.endsWith("extensions.xml")) {
+                            return "extensions";
+                        }
+                        return null;
+                    },
+                    // setting the values
+                    filePath -> filePath));
```
aneveux, line 84: That looks like an overly complex way of doing what you want to do.

Basically you want: 

* To search in `.mvn` for 2 files, `maven.config` and `extensions.xml`
* If they exist run some checks on them

Something like:

```java
try (Stream<Path> incrementalsConfigs = Files.find(mvnConfig, 1, (path, $) -> Files.isRegularFile(path) && (path.endsWith("maven.config") || path.endsWith("extensions.xml")))) {
```

Should directly give you a Stream that should contain 2 elements being the files you are searching for.

You can simply go for something like:

```java
Optional<Path> mavenExtensionsFile = incrementalsConfigs.filter(path -> path.endsWith("extensions.xml")).findFirst(); 

Optional<Path> mavenConfigFile = incrementalsConfigs.filter(path -> path.endsWith("maven.config")).findFirst(); 
```

Then `map` those `Optional` with whatever validation method you implemented, and if both result are present and correct, you've got a Probe Success =)

aneveux (commented):
Some small comments, but overall that is a very good job @Jagrutiti !

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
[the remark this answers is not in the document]
Jagrutiti: > The important thing to keep in mind is that logs should relate to your application's execution, not your application's business implementation.

This helped and so did the article. Thanks.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
Jagrutiti: > ```java
> Optional<Path> mavenExtensionsFile = incrementalsConfigs.filter(path -> path.endsWith("extensions.xml")).findFirst(); 
> 
> Optional<Path> mavenConfigFile = incrementalsConfigs.filter(path -> path.endsWith("maven.config")).findFirst(); 
> ```

I tried something like this initially. It throws an error saying `java.lang.IllegalStateException: stream has already been operated upon or closed`.

I read about it and learned that `Streams` are one-time use only. 

I implemented what you suggested and got the same error.  I can only execute the first statement. Not the second one :(

Jagrutiti: One of the solutions is to use `Supplier<Stream<Path>>` but then I cannot use it in `try` because it requires auto-closeable types.

Jagrutiti: I took a different approach and was able to handle `Supplier` this time.

aneveux: Oh yeah indeed, sorry I misled with code examples. The usual ways to tackle this are either to store things in Collection forms and generate new streams whenever needed, or indeed go for suppliers.

Suppliers seem like a good idea for your situation, and great job figuring out a solution!

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,147 @@
...
+        Supplier<Stream<Path>> incrementalConfigs = () -> {
+            try {
+                return Files.find(mvnConfig, 1, (path, $) -> Files.isRegularFile(path) && (path.endsWith("maven.config") || path.endsWith("extensions.xml")));
+            } catch (IOException e) {
+                LOGGER.error("Could not read files from .mvn directory for {} plugin while running {} probe.", plugin.getName(), key());
+                return null;
+            }
+        };
+        Optional<Path> mavenExtensionsFile = incrementalConfigs.get().filter(path -> path.endsWith("extensions.xml")).findFirst();
+        Optional<Path> mavenConfigFile = incrementalConfigs.get().filter(path -> path.endsWith("maven.config")).findFirst();
```
aneveux, line 81: Your code works, but I just wanted to share something here to ensure you understand the whole thing.

In that particular case, it is not a problem, but basically, that piece of code will explore the `mvnConfig` directory each time you will call the supplier. Not a big deal in our case because we do not expect that directory to contain a lot of files, but in other similar situations, this could be an issue, because there is no particular need of actually processing *x* times the same folder.

The problem you are trying to solve here is that a `Stream` can only be consumed once, and the Supplier approach allows you to retrieve a new `Stream` everytime you call the supplier. But you can actually implement a `Supplier` that will create for you a new `Stream` out of data which you already collected.

See for example:

```java
try (Stream<Path> incrementalConfigsStream = Files.find(mvnConfig, 1, (path, $) -> Files.isRegularFile(path) && (path.endsWith("maven.config") || path.endsWith("extensions.xml")))) {
    List<Path> incrementalConfigsStore = incrementalConfigsStream.toList();
    
    // Here, we stored the data in a `List`, and we can create a supplier that will create a new `Stream` for us from this data
    Supplier<Stream<String>> incrementalConfigs = () -> incrementalConfigsStore.stream();

    // And now you can use your `Supplier` to retrieve whatever you want, and the `mvnConfig` directory will be read just once
        Optional<Path> mavenExtensionsFile = incrementalConfigs.get().filter(path -> path.endsWith("extensions.xml")).findFirst();
        Optional<Path> mavenConfigFile = incrementalConfigs.get().filter(path -> path.endsWith("maven.config")).findFirst();

}
```

WDYT?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,147 @@
...
+            DocumentBuilder dBuilder = dbFactory.newDocumentBuilder();
+            Document doc = dBuilder.parse(path.toFile());
+            doc.getDocumentElement().normalize();
+            Element rootElement = doc.getDocumentElement();
+            Element extensionElement = (Element) rootElement.getElementsByTagName("extension").item(0);
+            Element groupIdElement = (Element) extensionElement.getElementsByTagName("groupId").item(0);
+            Element artifactIdElement = (Element) extensionElement.getElementsByTagName("artifactId").item(0);
+            String groupId = groupIdElement.getTextContent();
+            String artifactId = artifactIdElement.getTextContent();
+            return groupId.equals(INCREMENTAL_TOOL) && artifactId.equals(INCREMENTAL_TOOL_ARTIFACT_ID);
```
aneveux, line 124: Remember that trick Adrien shared with you some time ago?

If `groupId` or `artifactId` are `null`, then calling the `equals` method on them will result in a `NullPointerException`, which you can avoid by simply writing your tests in the other way around, starting with what you know for sure is not `null`:

`INCREMENTAL_TOOL.equals(groupId) && INCREMENTAL_TOOL_ARTIFACT_ID.equals(artifactId)`

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Checks whether `maven.config` is configured in the plugin
+     *
+     * @param path Looks for extensions.xml configuration in the particular path
+     * @return true if a correct configuration is found, otherwise false
+     */
+    private boolean isMavenConfigConfigured(Path path) {
+        try {
+            return Files.lines(path).toList().containsAll(List.of("-Pconsume-incrementals", "-Pmight-produce-incrementals"));
```
aneveux, line 141: I think instead of `lines().toList()` you can directly use `.readAllLines()` which returns a `List` and not a `Stream`.

aneveux (commented):
Again a very good job @Jagrutiti ! Keep up the great rhythm :muscle: 

Left just very few remarks about some design discussion.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
```diff
@@ -0,0 +1,151 @@
...
+            Optional<Path> mavenConfigFile = incrementalConfigs.get().filter(path -> path.endsWith("maven.config")).findFirst();
+
+            if (mavenExtensionsFile.isPresent() && mavenConfigFile.isPresent()) {
+                return isExtensionsXMLConfigured(mavenExtensionsFile.get()) && isMavenConfigConfigured(mavenConfigFile.get())
+                    ? ProbeResult.success(key(), String.format("Incremental Build is configured in the %s plugin.", plugin.getName()))
+                    : ProbeResult.failure(key(), String.format("Incremental Build is not configured in the %s plugin.", plugin.getName()));
+            }
+        } catch (IOException e) {
+            LOGGER.error("Could not read files from .mvn directory for {} plugin while running {} probe.", plugin.getName(), key());
+            return null;
```
aneveux, line 88: Is it expected that we return `null`? Isn't it a probe failure? I don't remember what we usually do in that scenario TBH

Jagrutiti: This is the part of the old syntax. When I was using the Supplier the other way. My bad.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
Jagrutiti: I will remember more.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
Jagrutiti: This is a much better of doing things. 

I do not completely understand `Functional Interface` and  `Supplier` that well. They still scare me.

aneveux: These are common things coming from Functional Programming. The concepts apply to lots of languages, not Java only.

You can have a look at this 101 guide from Cassidy Williams for example: https://github.com/readme/guides/functional-programming-basics

If you're still unsure and would like to discuss it a bit so we can share some examples live, LMK, we can have a quick call for more explanation.

aneveux (approved):
Great job @Jagrutiti !

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/IncrementalBuildDetectionProbe.java
Jagrutiti: Thanks for the link and offering for a quick meeting. I will let you know if the need arises.

alecharp (approved):
This looks good!

PR #359, opened 2023-08-27
JSR-305 probe. Checks for `@Nonnull` and `@CheckForNull` deprecated annotations.

### LINKED ISSUE #322: `@Nonnull` and `@CheckForNull` replacement
### Description

Annotations `@Nonnull` and `@CheckForNull` (part of JSR-305) have been [deprecated since 2016](https://jcp.org/en/jsr/detail?id=305).
They should be replaced with SpotBugs annotation.

More detailed description can be find at https://www.jenkins.io/doc/developer/tutorial-improve/replace-jsr-305-annotations/.

Jagrutiti:
This probe checks for deprecated imports of `@Nonnull` and `@CheckForNull` annotations. 

### Description

Closes #322 

### Testing done
Added required test cases to validate the new probe. 

```[tasklist]
### Submitter checklist
```


core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
```diff
@@ -0,0 +1,83 @@
...
+import io.jenkins.pluginhealth.scoring.model.ProbeResult;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * The abstract probe checks for deprecated imports in a plugin.
+ */
+public abstract class AbstractDeprecatedAnnotationImportsCheckProbe extends Probe {
```
Jagrutiti, line 23: I was doubtful whether I should go for an abstract class. 

Though I do not know all the problems that come with plugin maintenance, it seems that in the future a situation may arise where we might need to search in `.java` files.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @return a failure message.
+     */
+    abstract String getFailureMessage();
+
+    /**
+     * @return a list of deprecated imports.
+     */
+    public abstract List<String> getListOfDeprecatedImports();
```
Jagrutiti, line 82: I made this a list so that multiple lines that need to be searched for could be added. Instead of hard-coding this and restricting it to a couple of files.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
Jagrutiti: I hope this isn't an incorrect decision. The goal is to make it future proof but if you think this situation may not arise, I am open to refactoring it.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JSR305Probe.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+/**
+ * This probe checks for deprecated @Nonnull and @CheckForNull annotations in a plugin.
+ */
+@Component
+@Order(value = AbstractDeprecatedAnnotationImportsCheckProbe.ORDER)
+public class JSR305Probe extends AbstractDeprecatedAnnotationImportsCheckProbe {
```
Jagrutiti, line 13: Here I was not able to think a good name for the probe. So I went ahead with this one.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
aneveux: Without entering the debate about whether we should keep it or not, I just wanted to share some interesting readings about some principles we (should?) tend to follow while writing code: https://blog.devgenius.io/key-principles-in-software-dry-kiss-yagni-solid-and-other-acronyms-98e5575a6942?gi=09e171025fdc

The key take aways in the current situation are *KISS* and *YAGNI*.

*KISS* (Keep It Simple, Stupid) is all about trying to write code in the simplest way possible.

*YAGNI* (You Ain't Gonne Need It) is slightly related and is all about wondering if you're really ever going to need what you implemented.

Again, I am not saying we should or should not go for that design you implemented, but rather sharing some general tips about coding: in general, we try to stick to the minimum change we can do in order to implement the feature that we need. Preparing stuff for a potential future is interesting only if we actually did study a bit what is about to come and we know for sure that we will need that preparation.

OTOH though, I really do appreciate both the effort, and the comments explaining your rationale. 

The decision itself is neither correct or incorrect, this is a Pull Request, so you are proposing something, with a rationale. Then we discuss it all together as a team working on the project, and we all agree on what seems to be the best decision based on everyone's arguments.

Great move on triggering that discussion Jagruti!

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * The abstract probe checks for deprecated imports in a plugin.
+ */
+public abstract class AbstractDeprecatedAnnotationImportsCheckProbe extends Probe {
+    public static final int ORDER = SCMLinkValidationProbe.ORDER;
```
aneveux, line 24: Is it supposed to run at the same time as `SCMLinkValidationProbe` or after it? I believe we need the source code to be available before checking it, so we might need to run after this probe?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractDeprecatedAnnotationImportsCheckProbe.class);
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        final Path scmRepository = context.getScmRepository();
+
+        /* The "maxDepth" is set to Integer.MAX_VALUE because a repository can have multiple modules with class files in it. We do not want to miss any ".java" file.  */
```
aneveux, line 31: Awesome work writing comments like that! Very good practice!

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
```diff
@@ -0,0 +1,83 @@
...
+        try (Stream<Path> paths = Files.find(scmRepository, Integer.MAX_VALUE, (path, $) ->
+            Files.isRegularFile(path) && path.getFileName().toString().endsWith(".java"))) {
```
aneveux, line 33: I believe you can use `endsWith` directly on a `Path` object without having to transform it in a `String`.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
```diff
@@ -0,0 +1,83 @@
...
+            Set<String> deprecatedPaths = paths.map(this::getFileNameWithDeprecatedImport)
```
aneveux, line 34: Not really `deprecatedPaths` right? But rather something like `javaFilesWithDeprecatedImports`?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Looks for all the deprecated imports in a path.
+     *
+     * @param path The file Path to check deprecated imports.
+     * @return name of the file that contains deprecated imports.
+     */
+    private String getFileNameWithDeprecatedImport(Path path) {
+        try (Stream<String> fileContent = Files.lines(path)) {
```
aneveux, line 59: Something interesting to think about in here: 

We only care about the `import` statements for that probe. So you could `filter` the lines of the file to retrieve only the one starting with `import`. What do you think?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @return a failure message.
+     */
+    abstract String getFailureMessage();
+
+    /**
+     * @return a list of deprecated imports.
+     */
+    abstract List<String> getListOfDeprecatedImports();
```
aneveux, line 82: While we are at abstracting code, you can see that your abstract probe is actually not just allowing implementations to search for deprecated imports, but for any imports... Which opens the way for more probes to be implemented on top of that.

I believe if we keep this abstract probe, it should actually be called `AbstractJavaImportsCheckProbe` because that is what it is actually doing. It does not look just for annotation imports, or deprecated imports, but you can make it search for any imports.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JSR305Probe.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "The probe checks for deprecated annotations.";
+    }
+
+    @Override
+    String getSuccessMessage() {
+        return "JSR305 annotation is updated";
```
aneveux, line 28: Or it is just not found right?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JSR305Probe.java
aneveux: lgtm

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JSR305Probe.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    String getFailureMessage() {
+        return "Deprecated annotations found";
+    }
+
+    @Override
+    public List<String> getListOfDeprecatedImports() {
+        return List.of(
+            "import javax.annotation.Nonnull;",
+            "import javax.annotation.CheckForNull;"
```
aneveux, line 40: If you're going for an implementation based on a `abstract` class that is supposed to deal with imports, you should not have to write the full `import` statement in here.

Just writing the fully qualified name of the import you are interested in, like `javax.annotation.Nonnull` should be enough, and it is up to the abstract class to then create the proper line with `import xxx;`.

(think that if you had to write another probe based on that, you would have to repeat writing `import xxx;` which is error prone and has no particular value).

aneveux (commented):
Left some comments, but overall that is a very good job @Jagrutiti ! 

Very much appreciate the comments you left for sharing your ideas, rationale, etc.

Great contribution!

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JSR305Probe.java
Jagrutiti: I did not use the word "not found" here because it gives a negative connotation.

Like, the deprecated annotations should be found but they are not. 

But this probe is successful when they are not found.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
Jagrutiti: Thanks for your encouraging words. 

I read the article and according to the _YAGNI_ the abstract shouldn't be a part as of now. 

>  Then we discuss it all together as a team working on the project, and we all agree on what seems to be the best decision based on everyone's arguments.

I agree with this. Whatever the decision is, it will be fine by me.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
Jagrutiti: In the first line: ` context.getScmRepository();` I am fetching the repository link. I want the link to be validated before I fetch the repo. 

So the `SCMLinkValidationProbe` should be executed first. I have also added it in the `getProbeResultRequirement`

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
Jagrutiti: This honestly makes so much sense. Thanks for pointing out.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
aneveux: Sorry, what I meant is: is this OK to use the same value for `ORDER` or should we increment it?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/JSR305Probe.java
aneveux: What I mean is perhaps JSR305 was never used in the plugin in the first place, so reading a message saying it was updated could be confusing.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractDeprecatedAnnotationImportsCheckProbe.java
Jagrutiti: https://javatechonline.medium.com/order-annotation-in-spring-boot-499d76c5a29c

In this article, it says:

> If Component does not have @Order annotation, it will execute in alphabetical order of bean name by default. Furthermore, if two components are having the same order value, the execution will happen in the alphabetical order of component name.

You are right. I added + 100 in that case. So that this probe run after `SCMLinkValidationProbe`.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractJavaImportsCheckProbe.java
```diff
@@ -0,0 +1,93 @@
...
+public abstract class AbstractJavaImportsCheckProbe extends Probe {
+    public static final int ORDER = SCMLinkValidationProbe.ORDER + 100;
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractJavaImportsCheckProbe.class);
+
+    @Override
+    protected ProbeResult doApply(Plugin plugin, ProbeContext context) {
+        final Path scmRepository = context.getScmRepository();
+
+        /* The "maxDepth" is set to Integer.MAX_VALUE because a repository can have multiple modules with class files in it. We do not want to miss any ".java" file.  */
+        try (Stream<Path> paths = Files.find(scmRepository, Integer.MAX_VALUE, (path, $) ->
```
aneveux, line 32: Here `paths` is `javaFiles` right? Or `javaFilesPath`?

The difference to me with something like `(path, $) ->` that you see at the end of the line is that at the end of the line `path` is indeed a path, and we don't know what it is exactly, we just know it's a path.

But the `Stream` that we create definitely contains all the Java files of the plugin, so we can have a more explicit name than just `paths`, because we know what kind of path that is.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractJavaImportsCheckProbe.java
```diff
@@ -0,0 +1,93 @@
...
+            Files.isRegularFile(path) && path.getFileName().toString().endsWith(".java"))) {
```
aneveux, line 33:
```suggestion
            Files.isRegularFile(path) && path.getFileName().endsWith(".java"))) {
```

No need for the `toString()` method IIRC.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractJavaImportsCheckProbe.java
```diff
@@ -0,0 +1,93 @@
...
+            Set<String> javaFilesWithDeprecatedImports = paths.map(this::getFileNameWithImports)
```
aneveux, line 34: Now that you changed the abstraction of your class to "detection of deprecated imports" to actually "detection of imports", you can rename `javaFilesWithDeprecatedImports` to what it became, something like: `javaFilesWithDetectedImports` or something similar.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractJavaImportsCheckProbe.java
```diff
@@ -0,0 +1,93 @@
...
+                .filter(Objects::nonNull)
```
aneveux, line 35: From a design perspective, I think you can slightly change the methods you are using.

Here, your `getFileNameWithImports` is transparently responsible of 2 different things:

* Checking if a file contains some `import` statements,
* Return the name of the file if it contains the `import` statements.

In order to implement those 2 behaviors in only 1 method, you opted for some sort of convention: if the method returns a `String`, it means the file contains the import, and you now have its file name. But if the file does not contain the import, you return `null` and that `null` value is to be interpreted in "the import statement is not found in that file".

If you remember the article I shared with you some time ago about accronyms, there is one related to implementing methods which are responsible for one and only one thing.

Here, your method is doing 2 different things.

One way to handle this could be:

Change `getFileNameWithImports` to a method like `boolean containsImports(Path javaFile, List<String> imports)`. The code you wrote is correct, you just have to change the return statement to a boolean instead of returning `String`s.

Then, in your caller method, instead of implementing that "convention" of yours (filtering the `null` objects to remove the java classes which do not contain the imports (see how non intuitive that is?)), you can simply do the following:

`javaFiles.filter(javaFile -> containsImports(javaFile, getListOfImports()))`

This first part will simply filter the list of all java files, keeping only the ones with the detected imports.

Then, because you want to collect their file names, you just have to:

`.map(javaFile -> javaFile.getFileName().toString())`.

This way you don't rely on kind of an internal transparent convention to write your code, and your code becomes very explicit.

WDYT? Does it make sense?

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractJavaImportsCheckProbe.java
```diff
@@ -0,0 +1,93 @@
...
+    abstract String getFailureMessage();
+
+    /**
+     * Gets a fully qualified name of the imported class/library from an {@code import} statement.
+     *
+     * @param importStatement the statement from which the fully qualified name should be fetched.
+     * @return a String that contains only the fully qualified name of the class/library.
+     */
+    private String getFullyQualifiedImportName(String importStatement) {
+        return importStatement.replace("import ", "").replace(";", "");
```
aneveux, line 86: You can add a `trim()` to be sure that there are no particular spaces remaining around the statement that you want.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/AbstractJavaImportsCheckProbe.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Looks for all the imports in a path that should be checked.
+     *
+     * @param path The file Path to check imports.
+     * @return name of the file that contains the required imports.
+     */
+    private String getFileNameWithImports(Path path) {
+        try (Stream<String> fileContent = Files.lines(path).filter(line -> line.startsWith("import")).map(this::getFullyQualifiedImportName)) {
```
aneveux, line 59: Notice that now that you refactored your code, `fileContent` is not really the file content? It became the `importStatements` :wink:

aneveux (commented):
You are doing a great job @Jagrutiti ! Really happy about this PR!

Left a bunch of comments to go a step further, please let me know if that does not make sense to you or if you think those are not worthy comments.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JSR305ProbeTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Test
+    void shouldNotReturnPluginsWithUpdatedAnnotations() throws IOException {
+        final Path repo = Files.createTempDirectory("foo");
+        Path directory = Files.createDirectories(repo.resolve("src/main/java"));
+        final Path javaFileWithUpdatedAnnotation = Files.createFile(directory.resolve("test-class-3.java"));
+        Files.createFile(directory.resolve("test-dummy-class-should-not-be-returned.java"));
+
+        Files.write(javaFileWithUpdatedAnnotation, List.of(
+            "package test;"
+        ));
```
aneveux, line 101: Actually, the updated annotations are: `edu.umd.cs.findbugs.annotations.CheckForNull` and `edu.umd.cs.findbugs.annotations.NonNull` if you want to actually check for "updated annotations".

aneveux (approved):
That is a really good work, well done @Jagrutiti !

zbynek:
To keep this compatible with the [OpenRewrite rule](https://github.com/openrewrite/rewrite-jenkins/blob/98bfbecc9c9507a069c9ac1f1e35cd5dd0fa0687/src/main/resources/META-INF/rewrite/rewrite.yml) by @sghill , it would be better to look for all imports from the `javax.annotation` package.

Jagrutiti:
> To keep this compatible with the [OpenRewrite rule](https://github.com/openrewrite/rewrite-jenkins/blob/98bfbecc9c9507a069c9ac1f1e35cd5dd0fa0687/src/main/resources/META-INF/rewrite/rewrite.yml) by @sghill , it would be better to look for all imports from the `javax.annotation` package.

Do you mean I should look for any import line that starts with `javax.annotation`?

zbynek:
Yes, that's what I meant.

aneveux:
@Jagrutiti Please, you can keep your abstract class to check for multiple imports (the way you had it before) and just change your JSR305 probe to have an unique entry in the list.

Or you remove your abstract class altogether and have a single one that does JSR305 and no abstraction.

Now that you initially implemented an abstract probe that allows searching for multiple imports a file, I would say you can keep it as is, and in the really context of JSR305, you just happen to be searching for a single import.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JSR305ProbeTest.java
```diff
@@ -0,0 +1,114 @@
...
+        assertThat(probe.apply(plugin, ctx))
+            .usingRecursiveComparison()
+            .comparingOnlyFields("id", "message", "status")
+            .isEqualTo(ProbeResult.failure(JSR305Probe.KEY, "Deprecated imports found at foo plugin for test-class-1.java, test-class-2.java, test-class-3.java"));
+        verify(probe).doApply(any(Plugin.class), any(ProbeContext.class));
+
+    }
+
+    @Test
+    void shouldNotReturnPluginsWithUpdatedAnnotations() throws IOException {
```
alecharp, line 93:
```suggestion
    void shouldNotReturnPluginsWithNoDeprecatedImports() throws IOException {
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JSR305ProbeTest.java
alecharp: I agree, this is not really what you are testing. You are "just" validating that the classes have no deprecated imports. 
It would be great to catch updated imports as well, so we could score (later on) plugins with updated imports better than plugin with no imports which would be scored better than plugins with deprecated imports (updated > normal > deprecated).

alecharp (commented):
I feel this is a good start but could benefit from checking more than just not seeing invalid imports, but also checking that new imports are there. WDYT @aneveux?

zbynek:
@alecharp 

> updated > normal 

that might punish plugins that use different means to prevent NPEs (e.g. using `Optional<>` consistently, having extensive test coverage, using another framework like jspecify ...)

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JSR305ProbeTest.java
aneveux: That's kinda tricky though cause we would need to define if it's legit not too use the annotations?

aneveux:
I agree with @zbynek on that one. The goal is to avoid using outdated annotations, but there are many ways to properly handle null values. Attributing good marks to plugins using the updated annotations for null management would lead to plugins having less good marks despite maybe not having any flaws.

I feel that while it is factual to check if a plugin is using deprecated annotations or not, deciding if the plugin is properly handling NPEs is much more opinionated, and I wouldn't recommend handling that in probes (we could assume poor NPE management will result in issues and bug reports, which are checked by other probes).

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JSR305ProbeTest.java
alecharp: true. Let's keep the "if used, they need to be the new ones" approach.

aneveux: The latter: 

> just test that when the new imports are used the probe does not fail?

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JSR305ProbeTest.java
```diff
@@ -0,0 +1,117 @@
...
+        final Path repo = Files.createTempDirectory("foo");
+        Path directory = Files.createDirectories(repo.resolve("src/main/java"));
+        final Path javaFileWithUpdatedAnnotation = Files.createFile(directory.resolve("test-class-3.java"));
+        Files.createFile(directory.resolve("test-dummy-class-should-not-be-returned.java"));
+
+        Files.write(javaFileWithUpdatedAnnotation, List.of(
+            "package test;",
+            "",
+            "edu.umd.cs.findbugs.annotations.NonNull",
+            "edu.umd.cs.findbugs.annotations.CheckForNull"
```
alecharp, line 103:
```suggestion
            "import edu.umd.cs.findbugs.annotations.NonNull;",
            "import edu.umd.cs.findbugs.annotations.CheckForNull;",
            "",
            "import java.util.HashMap;",
            "import java.util.Map;"
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/JSR305ProbeTest.java
```diff
@@ -0,0 +1,117 @@
...
+        final Path javaFileWithCheckForNullDeprecatedAnnotation = Files.createFile(directory.resolve("test-class-3.java"));
+        final Path javaFileWithNoDeprecatedAnnotation = Files.createFile(directory.resolve("test-class-4.java"));
+        final Path txtFileShouldNotBeFound = Files.createFile(directory.resolve("file.txt"));
+        Files.createFile(directory.resolve("test-dummy-class-should-not-be-returned.java"));
+
+        Files.write(javaFileWithAllDeprecatedAnnotation, List.of(
+            "package test;",
+            "",
+            "import javax.annotation.Nonnull;",
+            "import javax.annotation.CheckForNull;"
```
alecharp, line 58:
```suggestion
            "import javax.annotation.CheckForNull;",
            "",
            "import java.util.HashMap;",
            "import java.util.Map;"
```

alecharp (commented):
To make sure we are safe from false alarms.

alecharp (approved):
I'm satisfied with the content of the pull request. Thank you @Jagrutiti.

PR #372, opened 2023-10-04
feat: provide an ARM64 version of the image

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3619

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -60,7 +60,7 @@ pipeline {
 
     stage('Docker image') {
       steps {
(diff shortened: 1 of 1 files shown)


alecharp (approved):
I don't see a problem with it but wonder if it will have limitation at runtime.
We can always built it and see.

dduportal:
> I don't see a problem with it but wonder if it will have limitation at runtime. We can always built it and see.

Unless PHS uses a native librarythrough a JNI call, there is no difference between both CPU archs with a Java app ;)

dduportal (approved):

PR #392, opened 2023-10-24
Simplifies adoption scoring based on commit and release gap

alecharp:
### Description

This is to improve the outcome of the adoption scoring.
Currently, the adoption is evaluating the time gap between the last release of the plugin and its last commit date. 
However, all the milestone gap are evaluated even when a smaller gap was evaluated positively. 

### Testing done

Kept the existing test working.

```[tasklist]
### Submitter checklist
- [ ] If an issue exists, it is well described and linked in the description
```

### DIFF
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/scores/AdoptionScoring.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/scores/AdoptionScoring.java
@@ -46,7 +46,7 @@ public class AdoptionScoring extends Scoring {
     private static final String KEY = "adoption";
 
     private abstract static class TimeSinceLastCommitScoringComponent implements ScoringComponent {
-        public final Duration getTimeBetweenLastCommitAndDate(String lastCommitDateMessage, ZonedDateTime then) {
+        protected final Duration getTimeBetweenLastCommitAndDate(String lastCommitDateMessage, ZonedDateTime then) {
             final ZonedDateTime commitDate = ZonedDateTime
                 .parse(lastCommitDateMessage, DateTimeFormatter.ISO_DATE_TIME)
                 .withZoneSameInstant(getZone());
@@ -96,27 +96,7 @@ public class AdoptionScoring extends Scoring {
             new TimeSinceLastCommitScoringComponent() {
                 @Override
                 public String getDescription() {
-                    return "There must be less than 6 months between the last commit and the last release.";
-                }
-
-                @Override
-                public ScoringComponentResult getScore(Plugin plugin, Map<String, ProbeResult> probeResults) {
-                    final ProbeResult probeResult = probeResults.get(LastCommitDateProbe.KEY);
-                    if (probeResult == null || ProbeResult.Status.ERROR.equals(probeResult.status())) {
-                        return new ScoringComponentResult(-100, 100, List.of("Cannot determine the last commit date."));
-                    }
-
-                    final long days = getTimeBetweenLastCommitAndDate(probeResult.message(), plugin.getReleaseTimestamp().withZoneSameInstant(getZone())).toDays();
-                    if (days <= Duration.of(6 * 30, ChronoUnit.DAYS).toDays()) {
-                        return new ScoringComponentResult(100, getWeight(), List.of("There is less than 6 months between the last release and the last commit."));
-                    }
-                    return new ScoringComponentResult(0, getWeight(), List.of("There is more than 6 months between the last release and the last release."));
-                }
-            },
-            new TimeSinceLastCommitScoringComponent() {
-                @Override
-                public String getDescription() {
-                    return "There must be between 6 months and 1 year between the last commit and the last release.";
+                    return "There must be a reasonable time gap between last release and last commit.";
                 }
 
                 @Override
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/jenkins/pluginhealth/scoring/scores/AdoptionScoring.java
```diff
@@ -126,73 +106,23 @@ public ScoringComponentResult getScore(Plugin plugin, Map<String, ProbeResult> p
...
-                public String getDescription() {
-                    return "There must be a commit between 2 years and 4 years since the last release.";
-                }
-
-                @Override
-                public ScoringComponentResult getScore(Plugin plugin, Map<String, ProbeResult> probeResults) {
-                    final ProbeResult probeResult = probeResults.get(LastCommitDateProbe.KEY);
-                    if (probeResult == null || ProbeResult.Status.ERROR.equals(probeResult.status())) {
-                        return new ScoringComponentResult(-100, 100, List.of("Cannot determine the last commit date."));
+                    if (days < Duration.of(30 * 12 + 1, ChronoUnit.DAYS).toDays()) {
```
zbynek, line 113: The date computations here are strange.  First the conversion is done using a simple formula like `months *30` or `years*365`, then it's passed to `ChronoUnit` conversion from days to days. Would be nice to either do the conversion with `ChronoUnit.MONTHS` or just use the multiplication, but not both.

core/src/main/java/io/jenkins/pluginhealth/scoring/scores/AdoptionScoring.java
```diff
@@ -126,73 +106,23 @@ public ScoringComponentResult getScore(Plugin plugin, Map<String, ProbeResult> p
...
-                @Override
-                public ScoringComponentResult getScore(Plugin plugin, Map<String, ProbeResult> probeResults) {
-                    final ProbeResult probeResult = probeResults.get(LastCommitDateProbe.KEY);
-                    if (probeResult == null || ProbeResult.Status.ERROR.equals(probeResult.status())) {
-                        return new ScoringComponentResult(-100, 100, List.of("Cannot determine the last commit date."));
+                    if (days < Duration.of(30 * 48 + 1, ChronoUnit.DAYS).toDays()) {
+                        return new ScoringComponentResult(10, 2, List.of(defaultReason, "Less than 4 years between last release and last commit."));
                     }
-                    final long days = getTimeBetweenLastCommitAndDate(probeResult.message(), plugin.getReleaseTimestamp().withZoneSameInstant(getZone())).toDays();
                     if (days > Duration.of(4 * 365, ChronoUnit.DAYS).toDays()) {
```
zbynek, line 122: The check right above is for `days <  30*48+1`, so checking `days > 4 * 365` will return almost always true (unless you hit the interval between 1441 and 1460 days caused by discrepancy between the time conversions).

core/src/main/java/io/jenkins/pluginhealth/scoring/scores/AdoptionScoring.java
alecharp: I would love to do that. The problem is that any other `ChronoUnit` than `DAYS` is considered _estimated_ (see https://github.com/openjdk/jdk/blob/744e0893100d402b2b51762d57bcc2e99ab7fdcc/src/java.base/share/classes/java/time/temporal/ChronoUnit.java#L225-L228). Because of that, you get an exception in https://github.com/openjdk/jdk/blob/744e0893100d402b2b51762d57bcc2e99ab7fdcc/src/java.base/share/classes/java/time/Duration.java#L732-L734 (which is called by the [`ZERO.plus(amount, unit)`](https://github.com/openjdk/jdk/blob/744e0893100d402b2b51762d57bcc2e99ab7fdcc/src/java.base/share/classes/java/time/Duration.java#L311-L313), coming from [`Duration.of(amount, unit)`](https://github.com/openjdk/jdk/blob/744e0893100d402b2b51762d57bcc2e99ab7fdcc/src/java.base/share/classes/java/time/Duration.java#L311-L313)).

So, because anything greater than a day is "estimated". That's why all the duration here are in days.

core/src/main/java/io/jenkins/pluginhealth/scoring/scores/AdoptionScoring.java
alecharp: I changed that in https://github.com/jenkins-infra/plugin-health-scoring/pull/392/commits/b8b1238867ef8e5f2e4879b975a79f56d9ea5259

alecharp:
I like that idea @zbynek but I think we should tackle that in a different pull request.
It requires a different probe ("first-unreleased-commit" just an idea). 

Also, for now, yes a new maintainer could get a bad score on the plugin for a limited period of time, as once the plugin is released with a few pull requests, the problem would disappear. 
I would also argue that, to me, it would be better to release often when you get maintain a old plugin, as then it's possible to easily say "if you have a problem with version 1.2.3 but not with 1.2.2, it's because of PR 13".

core/src/main/java/io/jenkins/pluginhealth/scoring/scores/AdoptionScoring.java
```diff
@@ -126,73 +106,20 @@ public ScoringComponentResult getScore(Plugin plugin, Map<String, ProbeResult> p
...
+                        return new ScoringComponentResult(10, 2, List.of(defaultReason, "Less than 4 years between last release and last commit."));
                     }
-                    return new ScoringComponentResult(0, getWeight(), List.of("No commit in the last 4 years."));
-
-                }
-
-                @Override
-                public int getWeight() {
-                    return 0;
+                    return new ScoringComponentResult(-1000, getWeight(), List.of("There is more than 4 years between the last commit and the last release."));
```
zbynek, line 122:
```suggestion
                    return new ScoringComponentResult(-1000, getWeight(), List.of("There is more than 4 years between the last release and the last commit."));
```

zbynek:
> agree that this is a good enough fix

Yes 🚢. Just one minor suggestion for message wording.

PR #228, opened 2023-03-01
Add search, sort and pagination to probe listing page

lzeee:
Due to the unreasonable process of my previous PR, I opened this PR and kept the submission history of the original contributors. Thanks for your contribution, the valuable historical code review comments and the patient guidance! These led me to gradually learn how to contribute to the community. @dheerajodha @alecharp 

### History background
It's an old issue https://github.com/jenkins-infra/plugin-health-scoring/issues/65. What I have done is just trying to add the necessary packages to the system and make it work. **90% of the codes** comes from 2 historical PR** made by https://github.com/jenkins-infra/plugin-health-scoring/pull/109 @dheerajodha and https://github.com/jenkins-infra/plugin-health-scoring/pull/194 @impact-maker. 

### Description
This PR is an enhancement to the display page of the probe, allowing you to page, sort and search the probes. In order to accomplish this goal, I mainly did these things based on the historical code:
1. Add 3 npm packages(datatables.net、jquery、expose-loader) and make them work. Specify the '/probe' page when processing the table.
2. Add 1 *.less files to control the style of 4 added component, which is supplied by datatables.net.
3. Modify several probe classes to adapt to the newly added abstract function.

I think my code needs serious review, any suggestions and comments would be greatly appreciated !

the screen shot:
show

### Submitter checklist
- [ ] The description of this pull request is detailed and explain why this pull request is needed
- [ ] The changeset is on a specific branch
  - `feature/` for new feature, or improvements
  - `fix/` for bug fixes
  - `docs/` for any documentation changes
- [ ] If required, the documentation has been updated
- [ ] There is automated tests to cover the code change / addition
    - If there is no test, include an explanation why in the description
- [ ] Run `mvn verify` locally and all tests are passing successfully
  - It is OK to create a pull request which has failing tests if it is created as a draft, is to fix a bug and the first commit is the test to prove the existence of the bug.
- [ ] There is no new warnings (checkstyle nor spotbugs) on the code

PR #109, opened 2022-11-14
Adds search, sort, pagination feature to probes table

dheerajodha:
Fixes #65

This PR makes use of the [datatables](https://www.datatables.net/) tool to modify the table.
Untitled

PR #464, opened 2024-02-13
Add search, sort and pagination feature to probes table

### LINKED ISSUE #65: Sort probes in UI based on execution order or name
Because of 
https://github.com/jenkins-infra/plugin-health-scoring/blob/6f9c5d2363df9599117e310f88f4461f61249ba5/src/main/java/io/jenkins/pluginhealth/scoring/http/ProbesController.java#L55

the probes are listed based on their name. However, it might be good to sort based on their execution order on each plugin. Or to have and indication of the execution order and to be able to sort the list as the user wants.

sridamul:
Closes #65 

It's been quite a long time since @dheerajodha and @lzeee tried to resolve this issue. This is a continuation of these PRs.

As mentioned by the @alecharp in this [comment](https://github.com/jenkins-infra/plugin-health-scoring/pull/222#issuecomment-1450151243), I tried to fetch the branch from @dheerajodha's repo which ultimately failed.
`git@github.com: Permission denied (publickey).`
Also the branch is too old to work on. So, I started working in a new branch (but major props to @dheerajodha and @lzeee for their contributions)

This PR replaces [datatables](https://datatables.net/) with [simple-datatables](https://github.com/fiduswriter/simple-datatables) which primarily uses plain JS instead of jquery.

Made changes according to the reviews given by @alecharp in the previous PR's. But still I have few doubts in them.

- [Comment-1](https://github.com/jenkins-infra/plugin-health-scoring/pull/109#discussion_r1022497645) 👇
> We have a package manager for front-end libraries and a build system to package everything.
This is not a working solution to me.

From my understanding and previous implementation by @lzeee , I removed the CDN of simple-database and installed the npm package of it, then abstracted the css into a separate file 'simple-datatable.less'. Do I need to change anything else?

- [Comment-2](https://github.com/jenkins-infra/plugin-health-scoring/pull/228#discussion_r1122805024) 👇
> This simple complex. Can we not have a <script> in the correct template and not have this in the global javascript?

As mentioned by @lzeee, we can use 'html-webpack-plugin' I have also researched about the alternatives but this is the most preferrable one. Suggestions are welcomed. also why don't we keep datatable function in the index.js?

Currently the probes page look like this:

https://github.com/jenkins-infra/plugin-health-scoring/assets/107404972/44e90838-deee-4aaa-a6d2-7c45d8e3d0f6

### DIFF
--- a/war/package.json
+++ b/war/package.json
@@ -25,7 +25,7 @@
   },
   "dependencies": {
     "ionicons": "7.3.1",
-    "simple-datatables": "^9.0.0"
+    "simple-datatables": "9.0.3"
   },
   "browserslist": [
     "defaults",
--- a/war/package.json
+++ b/war/package.json
@@ -24,7 +24,8 @@
     "webpack-remove-empty-scripts": "1.0.4"
   },
   "dependencies": {
-    "ionicons": "7.3.1"
+    "ionicons": "7.3.1",
+    "simple-datatables": "^9.0.0"
   },
   "browserslist": [
     "defaults",
--- a/war/src/main/js/index.js
+++ b/war/src/main/js/index.js
@@ -1,9 +1,19 @@
 import {defineCustomElements} from "ionicons/dist/loader";
+import { DataTable } from "simple-datatables";
 
 defineCustomElements(window, {
   resourcesUrl: '/',
 });
 
+const probesTable = document.getElementById('probes-table');
+if (probesTable !== null) {
+    new DataTable("#probes-table", {
+      perPageSelect: [10, 25, 50],
+      columns: [
+        { select: 1, sortable: false },
+      ]
+    });
+}
 
 const updateCollapseIcon = (container, target) => {
   if (target.classList.contains('show')) {
--- a/war/src/main/less/index.less
+++ b/war/src/main/less/index.less
@@ -7,6 +7,7 @@
 @import './modules/page-footer';
 @import './modules/score';
 @import './modules/table';
+@import '~simple-datatables/src/css/style.css';
 
 html {
   font-size: 16px !important;
--- a/war/src/main/resources/templates/probes/listing.html
+++ b/war/src/main/resources/templates/probes/listing.html
@@ -12,7 +12,7 @@
         Raw results of the probes execution can be found on the
         <a href="" data-th-href="@{/probes/results}">results page</a>.
     </p>
-    <table class="table">
+    <table class="table" id="probes-table">
         <thead>
         <tr>
             <th>ID</th>
--- a/war/yarn.lock
+++ b/war/yarn.lock
@@ -1071,6 +1071,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"dayjs@npm:^1.11.10":
+  version: 1.11.10
+  resolution: "dayjs@npm:1.11.10"
+  checksum: 27e8f5bc01c0a76f36c656e62ab7f08c2e7b040b09e613cd4844abf03fb258e0350f0a83b02c887b84d771c1f11e092deda0beef8c6df2a1afbc3f6c1fade279
+  languageName: node
+  linkType: hard
+
 "del@npm:^4.1.1":
   version: 4.1.1
   resolution: "del@npm:4.1.1"
@@ -1083,6 +1090,13 @@ __metadata:
     pify: "npm:^4.0.1"
     rimraf: "npm:^2.6.3"
   checksum: 521f7da44bd79da841c06d573923d1f64f423aee8b8219c973478d3150ce1dcc024d03ad605929292adbff56d6448bca60d96dcdd2d8a53b46dbcb27e265c94b
+  languageName: node
+  linkType: hard
+
+"diff-dom@npm:^5.1.3":
+  version: 5.1.3
+  resolution: "diff-dom@npm:5.1.3"
+  checksum: c3eecb6a0487e49b8429474994d9236959805666699e9074cf9c9a7e39f6a11818a9b80a2c4af58ec74496c08bbae7eb53b64184817e79699fba1f8f21a2ffe5
   languageName: node
   linkType: hard
 
@@ -2160,6 +2174,7 @@ __metadata:
     path: "npm:0.12.7"
     postcss: "npm:8.4.38"
     postcss-loader: "npm:8.1.1"
+    simple-datatables: "npm:^9.0.0"
     style-loader: "npm:3.3.4"
     webpack: "npm:5.91.0"
     webpack-cli: "npm:5.1.4"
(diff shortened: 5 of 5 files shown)


core/src/main/java/io/jenkins/pluginhealth/scoring/probes/DependabotProbe.java
```diff
@@ -34,6 +34,7 @@
 @Order(AbstractDependencyBotConfigurationProbe.ORDER)
 public class DependabotProbe extends AbstractDependencyBotConfigurationProbe {
     public static final String KEY = "dependabot";
+    public static final int ORDER = AbstractDependencyBotConfigurationProbe.ORDER;
```
alecharp, line 37: In that case, make sure to use that `ORDER` in the `@Order` class annotation.

core/src/main/java/io/jenkins/pluginhealth/scoring/probes/RenovateProbe.java
```diff
@@ -10,6 +10,7 @@
 @Order(AbstractDependencyBotConfigurationProbe.ORDER)
 public class RenovateProbe extends AbstractDependencyBotConfigurationProbe {
     public static final String KEY = "renovate";
+    public static final int ORDER = AbstractDependencyBotConfigurationProbe.ORDER;
```
alecharp, line 13: Same.

sridamul:
If possible, how can I add 'dheerajodha' and 'lzeee' as contributors to this PR?

alecharp (changes requested):

alecharp:
> If possible, how can I add 'dheerajodha' and 'lzeee' as contributors to this PR?

For this, you needed to add 

```
Co-Authored-By: Username <email>
```

in your commit messages. See https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors.

It would be great to do that but it would mean to re-write your branch history and force push it. I'm usually against it but for this purpose, why not.

sridamul:
> > If possible, how can I add 'dheerajodha' and 'lzeee' as contributors to this PR?
> 
> For this, you needed to add 
> 
> ```
> Co-Authored-By: Username <email>
> ```
> 
> in your commit messages. See https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors.
> 
> It would be great to do that but it would mean to re-write your branch history and force push it. I'm usually against it but for this purpose, why not.

Right now I am moving back home, let me immediately do this after reaching home.

war/src/main/resources/templates/probes/listing.html
```diff
@@ -2,6 +2,7 @@
 <html lang="en" data-layout-decorate="~{layouts/main}">
 <head>
     <title>Probes</title>
+    <script src="" data-th-src="@{/js/index.js}" type="text/javascript"></script>
```
alecharp, line 5: this is already included in the `layouts/main` which this page is based on.

```suggestion
```

war/src/main/js/index.js
```diff
@@ -1,9 +1,11 @@
 import {defineCustomElements} from "ionicons/dist/loader";
+import { DataTable } from "simple-datatables";
 
 defineCustomElements(window, {
   resourcesUrl: '/',
 });
 
+new DataTable("#probes-table");
```
alecharp, line 8: this means that the `DataTable` is created on all pages, even when there is no div with the correct id.
Would it be possible to only activate it on the `probes/listing.html`?

alecharp (changes requested):

war/src/main/less/modules/simple-datatable.less
alecharp, line 1: I'm not sure why we need to have this entire file added in the project when it should come from the dependency directly.

sridamul: I thought it would be better to modify the design if we want in the future

alecharp: we shouldn't need the entire file but override the classes we want.

war/src/main/less/index.less
```diff
@@ -7,6 +7,7 @@
 @import './modules/page-footer';
 @import './modules/score';
 @import './modules/table';
+@import '/node_modules/simple-datatables/src/css/style.css';
```
alecharp, line 10:
```suggestion
@import '~simple-datatables/src/css/style.css';
```

sridamul:
Hey @alecharp, if possible, could you please check if any more changes are required?

alecharp:
@sridamul please, do not force-push on pull request that had already been reviewed. This means I have to check everything again.

alecharp:
@sridamul I'm sorry I pushed a few commits on your pull request. This is mostly to fix issues that are checked since Spotless setup and to finish the work you have done. 

Thank you for your contribution. I'm going to merge this shortly.

sridamul:
Hey @alecharp Is it ready to ship?

alecharp:
Currently no. There are plenty of conflicts.

sridamul:
Hey @alecharp I have resolved the conflicts and the changes seems to work fine as expected.
Following are the new changes made:
- added getOrder() fn in CodeOwnershipProbe
- applied spotbugs (which is the reason to change the copyright year and alignment in some probes)

alecharp:
> sorry for applying the spotless changes in single PR

that was required of you as you changed the probe codes. However, I'm not sure it was required to do so on the unit tests. I'm looking at it.

in the end, I like the sorting and search of the probes. But the Java code changes seems a bit too big for such a little feature. I'm not questioning the pull request, just the addition of the `getOrder` method. I'm wondering if just adding the UI feature without the order couldn't be enough.

sridamul:
> I'm not questioning the pull request, just the addition of the getOrder method. I'm wondering if just adding the UI feature without the order couldn't be enough.

do you mean to drop the execution order column?

alecharp:
> do you mean to drop the execution order column?

Yes.

sridamul:
Sorry for the force push!

Here is the final table lookup...

final_table

wdyt?

war/package.json
```diff
@@ -24,7 +24,8 @@
     "webpack-remove-empty-scripts": "1.0.4"
   },
   "dependencies": {
-    "ionicons": "7.3.1"
+    "ionicons": "7.3.1",
+    "simple-datatables": "^9.0.0"
```
alecharp, line 28:
```suggestion
    "simple-datatables": "9.0.0"
```

alecharp (changes requested):
> Sorry for the force push!

That's fine.

That would be great.
Could I ask of you to make sure to have the `9.0.3` version of `simple-datatables` and not use the `^` (which introduce version uncertainty). Thanks.

alecharp (approved):
Thank you very much for your time and effort on this pull request.

PR #472, opened 2024-02-22
Uses newest scoring timestamp as etag header

### LINKED ISSUE #373: Put in place a periodic export of scores to be consumed by plugin-site
> I was also thinking about putting in place a periodic export of plugin health scores like what's done in https://github.com/jenkins-infra/infra-reports so the plugin site could retrieve these scores from the same place instead of directly querying the plugin health score API. Then even if the plugin health score is down previous scores could be displayed.

> WDYT?

_Originally posted by @lemeurherve in https://github.com/jenkins-infra/plugin-site/issues/1337#issuecomment-1515957842_

alecharp:
### Description

Closes #373.

Not all scores are computed all the time. For a score to be computed again on a plugin, probe results on the plugin have to be updated (more recent than last execution of the score) and the scoring implementation must have the same version as on its last execution.

Because of that, we know that the most recent score computed, across all the plugins, will determine if the content of the `/scores/api` has changed since the last time it was fetched.

This puts the value of the seconds since epoch of the most recent scoring execution in the `ETag` header of the response.

### Testing done

Adjusted the unit test on the API to validate the presence and value of the `ETag` header.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ node_modules/
 ### IDE
 *.iml
 *.idea
+
+### fetch-report.sh
+etags.txt
+scores.json
--- a/Jenkinsfile_report
+++ b/Jenkinsfile_report
@@ -1,4 +1,4 @@
-def cronExpr = env.BRANCH_IS_PRIMARY ? 'H/30 * * * *' : ''
+def cronExpr = env.BRANCH_IS_PRIMARY ? 'H * * * *' : ''
 
 pipeline {
   agent any
--- a/Jenkinsfile_report
+++ b/Jenkinsfile_report
@@ -44,7 +44,7 @@ pipeline {
 
     stage('Publish') {
       when {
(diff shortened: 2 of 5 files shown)


lemeurherve (dismissed):

alecharp:
@lemeurherve is that enough or should we have something in front of the `/scores/api` interface to cache the response and "keep" the data as long as it's not changed?

lemeurherve:
I don't know.

lemeurherve:
> I definitively would be in favor of a static export of everything like it's done for other sources of data for plugin-site instead of an API, but that's a discussion not directly related to this PR.

Looking again at the linked issue, that's definitively related.

https://github.com/jenkins-infra/plugin-health-scoring/issues/373 is about having a static export instead of an API to pass data to plugin-site.

PR #556, opened 2024-10-15
chore(Jenkinsfile) avoid rebuilding tags and PRs daily

dduportal:
Fixup of https://github.com/jenkins-infra/plugin-health-scoring/pull/536 which introduces a daily cron trigger.

We forgot that this trigger should be "branch-conditionnal" otherwise it tries to rebuild everything including tags once a day (which is... not a good thing)

----

Notes:
- No need to trigger a release: this change will be picked by the jobs once merged
- Regarding the tags, I'll run a build replay to set them up to "never rebuild" with a successful status (dummy `echo OK` step) so it will stop them. Only the following tags are subject to the rebuild:
  - 4.0.0
  - 4.0.1
  - 4.0.2
  - 4.1.0

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 #!/usr/bin/env groovy
 
 // Do not rebuild daily if not on the principal branch (e.g. not on PR, not on other branches, not on tags)
-String cronPattern = env.BRANCH_IS_PRIMARY ? '@daily : ''
+String cronPattern = env.BRANCH_IS_PRIMARY ? '@daily' : ''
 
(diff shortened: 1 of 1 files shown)


alecharp (approved):

dduportal:
Good, thanks!

the tags mentioned in the body had been replayed with the following pipeline:

```
pipeline {
    agent none
    triggers {
        cron('')
    }
    stages {
        stage('Hello') {
            steps {
                echo 'OK'
            }
        }
    }
}
```

and now shows no more daily cron trigger!

PR #652, opened 2025-05-02
Only consider dependency bot pull request that are over 90days old

### LINKED ISSUE #651: Increase validity period for dependency pull request bots
Currently, as soon as one pull request from a dependency bot is detected, the scoring is reducing the score for the plugin.

Because it is recommended (when using other plugin as dependency) to use the BOM, and the BOM is released every week, it is easy for a plugin to see its score be degraded frequently.

It might be better to only reduce the score of a plugin when dependency pull request are not attended to. Per https://github.blog/security/supply-chain-security/a-smarter-quieter-dependabot/, dependabot is stopping after 90days without activities. We might want to follow this.

# Expected outcome

Only plugins with dependency pull request older than 90days have their score decreased.

darinpope:
### Description

Closes #651.

When you take a closer look, you'll see we are checking against the start of day 90 days ago. However, the tests aren't getting that close. I'm just checking for 91 days. I'm going to think about a few more tests, but this covers what I think we are looking for.

### Testing done

`mvn clean verify`

### Submitter checklist

### DIFF
--- a/core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
+++ b/core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
@@ -98,14 +98,9 @@ class DependabotPullRequestProbeTest extends AbstractProbeTest<DependabotPullReq
         final GHLabel dependenciesLabel = mock(GHLabel.class);
         when(dependenciesLabel.getName()).thenReturn("dependencies");
 
-        final Date thirtyDaysAgo = getThirtyDaysAgoAsDate();
-
         final GHPullRequest pr_1 = mock(GHPullRequest.class);
-        doReturn(thirtyDaysAgo).when(pr_1).getCreatedAt();
         final GHPullRequest pr_2 = mock(GHPullRequest.class);
-        doReturn(thirtyDaysAgo).when(pr_2).getCreatedAt();
         final GHPullRequest pr_3 = mock(GHPullRequest.class);
-        doReturn(thirtyDaysAgo).when(pr_3).getCreatedAt();
         when(ghRepository.getPullRequests(GHIssueState.OPEN)).thenReturn(List.of(pr_1, pr_2, pr_3));
 
         final DependabotPullRequestProbe probe = getSpy();
@@ -141,13 +136,16 @@ class DependabotPullRequestProbeTest extends AbstractProbeTest<DependabotPullReq
         when(pr_1.getLabels()).thenReturn(List.of(dependenciesLabel));
         doReturn(ninetyOneDaysAgo).when(pr_1).getCreatedAt();
         final GHPullRequest pr_2 = mock(GHPullRequest.class);
+        doReturn(List.of(dependenciesLabel)).when(pr_2).getLabels();
         doReturn(thirtyDaysAgo).when(pr_2).getCreatedAt();
         final GHPullRequest pr_3 = mock(GHPullRequest.class);
         when(pr_3.getLabels()).thenReturn(List.of(dependenciesLabel));
         doReturn(ninetyOneDaysAgo).when(pr_3).getCreatedAt();
         final GHPullRequest pr_4 = mock(GHPullRequest.class);
+        doReturn(List.of(dependenciesLabel)).when(pr_4).getLabels();
         doReturn(thirtyDaysAgo).when(pr_4).getCreatedAt();
         final GHPullRequest pr_5 = mock(GHPullRequest.class);
+        doReturn(List.of(dependenciesLabel)).when(pr_5).getLabels();
         doReturn(thirtyDaysAgo).when(pr_5).getCreatedAt();
         when(ghRepository.getPullRequests(GHIssueState.OPEN)).thenReturn(List.of(pr_1, pr_2, pr_3, pr_4, pr_5));
 
--- a/core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
+++ b/core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
@@ -25,6 +25,7 @@ package io.jenkins.pluginhealth.scoring.probes;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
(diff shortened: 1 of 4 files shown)


core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
```diff
@@ -93,19 +98,23 @@ void shouldAccessGitHubAPIAndSeeNoPullRequest() throws IOException {
         final GHLabel dependenciesLabel = mock(GHLabel.class);
         when(dependenciesLabel.getName()).thenReturn("dependencies");
 
+        final Date thirtyDaysAgo = getThirtyDaysAgoAsDate();
+
         final GHPullRequest pr_1 = mock(GHPullRequest.class);
+        doReturn(thirtyDaysAgo).when(pr_1).getCreatedAt();
```
alecharp, line 104: as the pull request doesn't have the label `dependencies`, it would have been filtered out first. We don't need to mock the `getCreatedAt` call. 

Or we can add the label `dependencies` on the pull request so that it's not removed from date consideration because it has the label.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
```diff
@@ -93,19 +98,23 @@ void shouldAccessGitHubAPIAndSeeNoPullRequest() throws IOException {
...
         final GHPullRequest pr_2 = mock(GHPullRequest.class);
+        doReturn(thirtyDaysAgo).when(pr_2).getCreatedAt();
```
alecharp, line 106: same.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
```diff
@@ -93,19 +98,23 @@ void shouldAccessGitHubAPIAndSeeNoPullRequest() throws IOException {
...
         final GHPullRequest pr_3 = mock(GHPullRequest.class);
-        when(ghRepository.getPullRequests(GHIssueState.OPEN)).thenReturn(
-            List.of(pr_1, pr_2, pr_3)
-        );
+        doReturn(thirtyDaysAgo).when(pr_3).getCreatedAt();
```
alecharp, line 108: same.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
```diff
@@ -93,19 +98,23 @@ void shouldAccessGitHubAPIAndSeeNoPullRequest() throws IOException {
         final GHLabel dependenciesLabel = mock(GHLabel.class);
         when(dependenciesLabel.getName()).thenReturn("dependencies");
 
+        final Date thirtyDaysAgo = getThirtyDaysAgoAsDate();
```
alecharp, line 101: not needed here.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
```diff
@@ -125,23 +134,30 @@ void shouldAccessGitHubAPIWhenDependabotActivatedWithOpenedPR() throws IOExcepti
...
+        final Date ninetyOneDaysAgo = getNinetyOneDaysAgoAsDate();
+
         final GHPullRequest pr_1 = mock(GHPullRequest.class);
         when(pr_1.getLabels()).thenReturn(List.of(dependenciesLabel));
+        doReturn(ninetyOneDaysAgo).when(pr_1).getCreatedAt();
         final GHPullRequest pr_2 = mock(GHPullRequest.class);
+        doReturn(thirtyDaysAgo).when(pr_2).getCreatedAt();
```
alecharp, line 144: To be needed, we need to have the label `dependencies` on the pull request.
```suggestion
        doReturn(List.of(dependenciesLabel)).when(pr_2).getLabels();
        doReturn(thirtyDaysAgo).when(pr_2).getCreatedAt();
```

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
```diff
@@ -125,23 +134,30 @@ void shouldAccessGitHubAPIWhenDependabotActivatedWithOpenedPR() throws IOExcepti
...
         final GHPullRequest pr_3 = mock(GHPullRequest.class);
         when(pr_3.getLabels()).thenReturn(List.of(dependenciesLabel));
+        doReturn(ninetyOneDaysAgo).when(pr_3).getCreatedAt();
         final GHPullRequest pr_4 = mock(GHPullRequest.class);
+        doReturn(thirtyDaysAgo).when(pr_4).getCreatedAt();
```
alecharp, line 149: same.

core/src/test/java/io/jenkins/pluginhealth/scoring/probes/DependabotPullRequestProbeTest.java
```diff
@@ -125,23 +134,30 @@ void shouldAccessGitHubAPIWhenDependabotActivatedWithOpenedPR() throws IOExcepti
...
         final GHPullRequest pr_5 = mock(GHPullRequest.class);
-        when(ghRepository.getPullRequests(GHIssueState.OPEN)).thenReturn(
-            List.of(pr_1, pr_2, pr_3, pr_4, pr_5)
-        );
+        doReturn(thirtyDaysAgo).when(pr_5).getCreatedAt();
```
alecharp, line 151: same.

alecharp (commented):
I do also have a problem locally running the tests. As discussed, I believed it's because of concurrent access of the method `getCreateAt()` which is a bridged method.
I'm investigating.

alecharp (changes requested):
I found out that the `doReturn|when` pattern is for spies, true but here it is used for mocks. 

As for the `WithBridgeMethod`, it introduces a method for binary compatibility but it does not make the annotated method use the generated method. 

Using

```
        when(pr.getCreatedAt()).thenReturn(someDaysAgo);
```

should be working without any problem.

darinpope:
I'm confused why you think that `when/thenReturn` should be working. I've made the other changes.

darinpope (commented):
I think I covered all the issues. Once I upgraded github-api to 2.0-rc.3, everything started working. However, I did not refactor to match the work being done in #569 since this is passing.

alecharp (approved):

Issue #317, opened 2023-05-26
Detects plugin using old parent pom version


alecharp:
### Description

Plugins are inheriting most of their build configuration through the parent pom (https://github.com/jenkinsci/plugin-pom). 
Recent modernization came from it.

It is a good practice to keep the parent pom up to date within any plugin configuration.
Detecting plugin using out of date parent pom could help guide maintainers or new contributors to participate on the project.

Jagrutiti:
Do I have to compare the version used in the child Pom inside the `<parent>` tag with this pom: https://github.com/jenkinsci/plugin-pom/blob/master/pom.xml?

In the case of Plugin Health Score, each module has its own pom. These children have a common super pom defined outside the modules. 

In this case, do I have to fetch the parent of the pom in each module and compare the version to the pom at https://github.com/jenkinsci/plugin-pom/blob/master/pom.xml?

AayushSaini101:
Please add GSOC label @alecharp

PR #779, opened 2026-01-03
feat: Add probe to detect parent POM version

rubrx:
### What this does

This PR adds a new probe that checks what parent POM version each Jenkins plugin is using. The probe reads the plugin's pom.xml and extracts the version number.

This is the first part of #317 - just collecting the data for now. We'll add the scoring/comparison part later once this is working.

### How it works

- Reads pom.xml from the plugin repository
- Extracts the `<parent><version>` using regex
- Returns the version number (like "4.80")
- Handles cases where pom.xml doesn't exist or has no parent

### Testing

I've tested this pretty thoroughly:
- 9 unit tests all passing
- Ran `mvn verify` successfully  
- Started the app locally and saw it running on actual plugins
- Confirmed error handling works (saw it handle missing pom.xml gracefully)

### Why Phase 1 only?

As @alecharp mentioned, the ideal solution needs a "scoring context" to fetch the latest plugin-pom version once and compare all plugins against it. That's architectural work that should be discussed separately.

This PR just collects the data - which is immediately useful and can be built upon later.

This is my first contribution here, so feedback is very welcome! Thanks for the guidance @alecharp. I am ready for review when you have time. Let me know if anything needs to be changed!

### DIFF
--- a/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ParentPomVersionProbe.java
+++ b/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ParentPomVersionProbe.java
@@ -1,0 +1,125 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2023-2026 Jenkins Infra
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/jenkins/pluginhealth/scoring/probes/ParentPomVersionProbe.java
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * <p>Jenkins plugins inherit build configuration from the parent POM
+ * (https://github.com/jenkinsci/plugin-pom). This probe extracts the
+ * parent POM version from the plugin's pom.xml to help track modernization
+ * and identify plugins using outdated parent versions.</p>
+ */
+@Component
+@Order(value = ParentPomVersionProbe.ORDER)
+public class ParentPomVersionProbe extends Probe {
+    public static final int ORDER = LastCommitDateProbe.ORDER + 100;
```
alecharp: I would suggest to do that after other Maven related probes.

alecharp (changes requested):
See the comments.

